PR #5, opened 2021-05-07
Homework

Nikber5:


Max-Pochepets (approved):
GJ

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElse(null);
```
KseniiaMakarova, line 21: returning `null` is bad practice - let's call `get()`

Nikber5: done

PR #8, opened 2021-05-11
solved dao-service task

freedmanMain:


mankokolya (approved):
I woul make names of query and statement shorter not createManufacturerStatement but createStatement same in all requests as it is undestandable from dao type what type you are working with. 
And set ) ENGINE=InnoDB AUTO_INCREMENT=4 to 1 because in your realization db will start to produce indexes from 4 but not 1

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,121 @@
...
+                PreparedStatement updateDriverStatement
+                        = connection.prepareStatement(updateDriverRequest)) {
+            updateDriverStatement.setString(1, driver.getName());
+            updateDriverStatement.setString(2, driver.getLicenseNumber());
+            updateDriverStatement.setLong(3, driver.getId());
+            if (updateDriverStatement.executeUpdate() > 0) {
+                return driver;
+            }
+            throw new NoSuchElementException(String.format(
+                    "Driver %s doesn't exist in DB", driver.toString()));
```
KseniiaMakarova, line 93:
```suggestion
                    "Driver %s doesn't exist in DB", driver));
```

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -7,6 +7,10 @@
     private String name;
     private String country;
 
+    public Manufacturer() {
+
```
KseniiaMakarova, line 11:
```suggestion
```

KseniiaMakarova (approved):

PR #9, opened 2021-05-11
added solution

Dalai-S:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private DriverDao driverDao;
+
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    public Driver get(Long id) {
+        return driverDao.get(id).orElse(null);
```
mankokolya, line 20: I would recommend you not to return null as it is a bad practice and you have to check in your controller(main) that you didn't get null. It is better to throw a runtime exception

mankokolya: You are not checking this in main

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,53 @@
...
+        Manufacturer audiManufacturer = new Manufacturer("Audi", "Germany");
+        Manufacturer infinityManufacturer = new Manufacturer("Infinity", "Japan");
+        Manufacturer teslaManufacturer = new Manufacturer("Tesla", "Tesla");
+
+        System.out.println(manufacturerService.create(audiManufacturer) + System.lineSeparator()
+                + manufacturerService.create(infinityManufacturer) + System.lineSeparator()
+                + manufacturerService.create(teslaManufacturer) + System.lineSeparator());
+
+        System.out.println(manufacturerService.get(16L) + System.lineSeparator()
+                + manufacturerService.get(999L));
```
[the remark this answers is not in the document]
Dalai-S: I can return null here, so the best way to show all info to avoid exception. I will do it through orElseThrow if the mentor says that what I wrote is bad.

src/main/java/mate/jdbc/dao/manufacturer/ManufacturerDaoImpl.java
```diff
@@ -39,7 +39,7 @@ public Manufacturer create(Manufacturer manufacturer) {
     @Override
     public Optional<Manufacturer> get(Long id) {
         String query = "SELECT * FROM manufacturers"
-                + " WHERE manufacturer_id = (?) AND deleted = FALSE";
+                + " WHERE id = (?) AND deleted = FALSE";
```
mankokolya, line 42: Do we need () around ?

src/main/java/mate/jdbc/Main.java
```diff
@@ -17,17 +19,17 @@ public static void main(String[] args) {
         Manufacturer infinityManufacturer = new Manufacturer("Infinity", "Japan");
         Manufacturer teslaManufacturer = new Manufacturer("Tesla", "Tesla");
 
-        System.out.println(manufacturerService.create(audiManufacturer) + System.lineSeparator()
-                + manufacturerService.create(infinityManufacturer) + System.lineSeparator()
-                + manufacturerService.create(teslaManufacturer) + System.lineSeparator());
+        System.out.println(manufacturerService.create(audiManufacturer));
```
mankokolya, line 22: take correct indexes from this value

mankokolya (changes requested):

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,65 @@
+package mate.jdbc.model;
+
+import java.util.Objects;
+
+public class Driver {
+    private Long id;
```
mankokolya, line 6: I don't see the file for creating database with tables

mankokolya (changes requested):

mankokolya (approved):

RomanDubovskyi (approved):
fix git conflicts

PR #32, opened 2021-05-16
Test hw

DanyloIT:
sorry for commit and pish from master(((((((((((((((((((((((((((((((((((((((((((

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,27 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+    private static final ManufacturerService manufacturerService =
+            (ManufacturerService) injector.getInstance(ManufacturerService.class);
+    private static final DriverService driverService =
+            (DriverService) injector.getInstance(DriverService.class);
+
     public static void main(String[] args) {
-        // test your code here
+        Driver bob = new Driver("Bob", "N765576");
+        Driver john = new Driver("John", "N342243");
+
+        System.out.println(driverService.create(bob));
+        System.out.println(driverService.create(john));
+        System.out.println(driverService.get(bob.getId()));
+        System.out.println(driverService.delete(john.getId()));
+        bob.setLicenseNumber("N765576P");
+        System.out.println(driverService.update(bob));
+        System.out.println(driverService.getAll());
     }
 }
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,27 @@
...
+        Driver bob = new Driver("Bob", "N765576");
+        Driver john = new Driver("John", "N342243");
+
+        System.out.println(driverService.create(bob));
+        System.out.println(driverService.create(john));
+        System.out.println(driverService.get(bob.getId()));
+        System.out.println(driverService.delete(john.getId()));
+        bob.setLicenseNumber("N765576P");
+        System.out.println(driverService.update(bob));
+        System.out.println(driverService.getAll());
```
Oleksandr-Shulha, line 25: Maybye better add manufacturerService commands also?

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE `drivers` (
+                           `id` bigint NOT NULL AUTO_INCREMENT,
+                           `name` varchar(255) COLLATE utf8_bin DEFAULT NULL,
+                           `license_number` varchar(255) COLLATE utf8_bin DEFAULT NULL,
+                           `is_deleted` tinyint NOT NULL DEFAULT '0',
+                           PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb3 COLLATE=utf8_bin;
```
Oleksandr-Shulha, line 7: what about Manufacterers table?

Oleksandr-Shulha (changes requested):

src/main/resources/init_db.sql
DanyloIT: agree, fixed

src/main/java/mate/jdbc/Main.java
DanyloIT: agree, fixed

Oleksandr-Shulha (approved):

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return Optional.of(driverDao.get(id).get());
```
Max-Pochepets, line 22: you should return value from service method, not Optional. but, even in your case, `driverDao.get(id)` already returns Optional, why you extract it and wrap again?

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/service/DriverServiceImpl.java
DanyloIT: fixed

Max-Pochepets (approved):

PR #33, opened 2021-05-16
First work edition

Oleksandr-Shulha:


src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
DanyloIT, line 21: In my Opinion its better to return Optional, but I am not 100% convinced.

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        String query = "SELECT * FROM drivers"
+                + " WHERE id = (?) AND deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement preparedStatement = connection.prepareStatement(query)) {
+            preparedStatement.setLong(1, id);
+            ResultSet resultSet = preparedStatement.executeQuery();
+            Driver driver = null;
```
DanyloIT, line 47: In my opinion, it is better to create and initialize driver in one line.

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id + " ",
+                    throwable);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long newId = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
+        String licenseNumber = resultSet.getString("license_number");
```
DanyloIT, line 109: Maybe we can insert this methods in constructor (in line 110)

DanyloIT (changes requested):

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
Oleksandr-Shulha: According "checklist" 
_find() methods in Dao layer should return Optional (not in services). In the service layer better call method get() on Optional and return the object._

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
Oleksandr-Shulha: what about resultSet.next()?

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
Oleksandr-Shulha: Ok. we can)

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
DanyloIT: agree

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
Oleksandr-Shulha: OK. Agree)

DanyloIT (approved):

RomanDubovskyi (approved):

PR #54, opened 2021-05-17
1ST ATTEMPT

yarrybka:


src/main/resources/init_db.sql
```diff
@@ -0,0 +1,19 @@
...
+
+
+CREATE TABLE `flights` (
+                           `id` int NOT NULL AUTO_INCREMENT,
+                           `number` varchar(10) DEFAULT NULL,
+                           `carrier` varchar(50) DEFAULT NULL,
+                           `manufacturer_id` int,
+                           `is_deleted` tinyint NOT NULL DEFAULT '0',
+                           PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```
LayRipper, line 19:
```suggestion
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;

```

LayRipper (approved):

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        return flight;
+    }
+
+    @Override
+    public Optional<Flight> get(Long id) {
+        String query = "SELECT * FROM flights WHERE id = ? AND is_deleted = FALSE;";
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(query)
+        ) {
```
RomanDubovskyi, line 47:
```suggestion
                 PreparedStatement statement = connection.prepareStatement(query)) {
```

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
```diff
@@ -0,0 +1,130 @@
...
+            statement.setString(1, flight.getNumber());
+            statement.setString(2, flight.getCarrier());
+            statement.setLong(3, flight.getManufacturer());
+            statement.setObject(4, flight.getId());
+            if (statement.executeUpdate() > 0) {
+                return flight;
+            } else {
```
RomanDubovskyi, line 89: we don't need else after return

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
```diff
@@ -0,0 +1,130 @@
...
+                throw new mate.jdbc.exception.DataProcessingException("Bad flight: " + flight);
```
RomanDubovskyi, line 90: why do we use package address here? 
and what does "bad flight" mean?

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
```diff
@@ -0,0 +1,130 @@
...
+                 PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete flight from DB by id = " + id, e);
+        }
+    }
+
+    @Override
+    public boolean truncateTable() {
```
RomanDubovskyi, line 111: we don't need this method since we use soft delete (is_deleted flag)

RomanDubovskyi (changes requested):
In this task we should've created Driver and model class. Flight is not equivalent to driver even if you want to create project in context of planes.

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
yarrybka: IDEA made it, when i was change orders of import. And I didn't this.

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
yarrybka: for example, when we will id, then bigger that we have in table

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
yarrybka: Can I left Flight, or change all to Pilote, for example?

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
yarrybka: Can I left Flight, or change all to Pilote, for example?

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
yarrybka: ok, it just help-method for make easier debbaging

src/main/java/mate/jdbc/dao/impl/FlightDaoImpl.java
RomanDubovskyi: I understand, but we don't need else block after we have return keyword or we throw exception in if. If we go inside of If code will stop execution after return or throwing exception.

src/main/java/mate/jdbc/dao/impl/PilotDaoImpl.java
```diff
@@ -0,0 +1,120 @@
...
+            statement.setString(2, pilot.getFlightNumber());
+            statement.setString(3, pilot.getCarrier());
+            statement.setLong(4, pilot.getManufacturer());
+            statement.setObject(5, pilot.getId());
+            if (statement.executeUpdate() > 0) {
+                return pilot;
+            } else {
+                throw new DataProcessingException("The pilot have illegal id or was deleted: "
+                        + pilot);
+            }
```
RomanDubovskyi, line 92:
```suggestion
            if (statement.executeUpdate() > 0) {
                return pilot;
                }
                throw new DataProcessingException("The pilot have illegal id or was deleted: "
                        + pilot);
```
same here

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,76 @@
...
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.model.Pilot;
+import mate.jdbc.service.ManufacturerService;
+import mate.jdbc.service.PilotService;
+
 public class Main {
+    private static final Injector INJECTOR = Injector.getInstance("mate.jdbc");
+    private static final ManufacturerService MANUFACTURER_SERVICE
```
RomanDubovskyi, line 11:
```suggestion
    private static final ManufacturerService manufacturerService
```
it's not a constant

RomanDubovskyi (approved):

PR #66, opened 2021-07-13
Implemented manufacturer and driver services and tested them in main …

deniskichan:
…method

### DIFF
--- a/src/main/java/mate/jdbc/service/DriverService.java
+++ b/src/main/java/mate/jdbc/service/DriverService.java
@@ -1,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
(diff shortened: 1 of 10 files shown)


SeReNb26 (approved):

src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
...
+    Driver create(Driver driver);
+
+    Driver get(Long id);
+
+    List<Driver> getAll();
+
+    Driver update(Driver driver);
+
+    boolean delete(Long id);
+
```
Sofasmile, line 16: let's remove empty line

Sofasmile (approved):
awesome

PR #87, opened 2021-09-14
implemented service for Manufacturer, created Driver model, implement…

AlexRozv:
…ed dao and service for Driver, added table config to init_db.sql, tested CRUD in main

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,41 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService = (ManufacturerService) injector
+                .getInstance(ManufacturerService.class);
+        DriverService driverService = (DriverService) injector
+                .getInstance(DriverService.class);
+        Manufacturer manufacturer = new Manufacturer("Honda", "Japan");
+        Driver driver = new Driver("Bob", "12345");
+        manufacturerService.create(manufacturer);
+        driverService.create(driver);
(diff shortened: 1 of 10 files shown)


Sinelife (changes requested):

Sinelife (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        Manufacturer manufacturer = new Manufacturer("Honda", "Japan");
+        Driver driver = new Driver("Bob", "12345");
+        manufacturerService.create(manufacturer);
+        driverService.create(driver);
+        List<Manufacturer> manufacturers = manufacturerService.getAll();
+        manufacturers.forEach(System.out::println);
+        List<Driver> drivers = driverService.getAll();
+        drivers.forEach(System.out::println);
+        Manufacturer updatedManufacturer = new Manufacturer("Toyota", "Japan");
+        updatedManufacturer.setId(1L);
```
Max-Pochepets, line 27: we can update and delete entities without setting ID. how can we do that?

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.lib.exception.DataProcessingException;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        String query = "INSERT INTO drivers (name, license_number) "
```
Max-Pochepets, line 20:
```suggestion
        String createDriverQuery = "INSERT INTO drivers (name, license_number) "
```

Max-Pochepets (changes requested):

src/main/java/mate/jdbc/Main.java
AlexRozv: fixed

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
AlexRozv: fixed

Max-Pochepets (approved):

PR #101, opened 2021-09-30
Jv dao servece solution

VladGural:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,67 @@
...
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Injector injector = Injector.getInstance("mate.jdbc");
+        final ManufacturerService manufacturerService = (ManufacturerService) injector
+                .getInstance(ManufacturerService.class);
+        final DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
```
Yelinskyi, line 15: You can move it to the class fields and make they private static.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,129 @@
...
+                  PreparedStatement getAllDriversStatement = connection
+                             .prepareStatement(getAllDriversRequest)) {
+            ResultSet resultSet = getAllDriversStatement.executeQuery();
+            while (resultSet.next()) {
+                allDriversList.add(parseResultSet(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get all Drivers from DB", e);
+        }
+        return allDriversList;
```
Yelinskyi, line 74: I'd moved return in try block.
But I don't know how is better.

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+import mate.jdbc.service.DriverService;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        List<Driver> drivers;
```
Yelinskyi, line 17: Note:
I think that you shoudn't save it in the collection if yo have a DB.
And at all a I understatnd I task we can use this class only for calling DAO methods.

Yelinskyi (changes requested):

Yelinskyi (approved):

Max-Pochepets (approved):

PR #104, opened 2021-10-01
JDBC_HW_03

AlexPravda1:
- Driver service and dao layers implemented
- Injector checked
- DB Connection checked

### DIFF
--- a/src/main/java/mate/jdbc/dao/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/DriverDaoImpl.java
@@ -1,0 +1,123 @@
+package mate.jdbc.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import lombok.extern.log4j.Log4j2;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.lib.exception.DataProcessingException;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+@Log4j2
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        log.info("DriverDao [create] method called for {}", driver);
+        String query = "INSERT INTO drivers (name, licenseNumber) "
+                + "VALUES (?, ?)";
+        try (Connection connection = ConnectionUtil.getConnection();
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+import mate.jdbc.lib.exception.DataProcessingException;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+@Log4j2
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        log.info("DriverDao [create] method called for {}", driver);
```
RomanDubovskyi, line 22: for future, we can skip logging now in this project

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        log.info("DriverDao [get] method was called for id {}", id);
+        String query = "SELECT * FROM drivers"
+                + " WHERE id = (?) AND is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getDriverStatement = connection.prepareStatement(query)) {
+            getDriverStatement.setLong(1, id);
+            ResultSet resultSet = getDriverStatement.executeQuery();
+            if (resultSet.next()) {
+                Driver driver = getDriver(resultSet);
+                return Optional.ofNullable(driver);
```
RomanDubovskyi, line 53:
```suggestion
                return Optional.of(driver);
```
it can't be null, so let'd use method of

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.List;
+import lombok.extern.log4j.Log4j2;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+@Log4j2
```
RomanDubovskyi, line 11: we don't log anything here

RomanDubovskyi (changes requested):

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
RomanDubovskyi: because your method getDriver can't return null

RomanDubovskyi (approved):

PR #119, opened 2021-10-04
implement driverDao, driverService and manufacturerService

andrii-tyndiuk:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
+    }
```
andrii-demchenko, line 23: Don't know if it is ok to implement get method like that. Not sure about that. And same in ManufacturerService. But let's see what mentors say.

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,17 @@
-CREATE SCHEMA IF NOT EXISTS `taxi` DEFAULT CHARACTER SET utf8;
-USE `taxi`;
-
+CREATE DATABASE `library_db` /*!40100 DEFAULT CHARACTER SET utf8 */ /*!80016 DEFAULT ENCRYPTION='N' */;
+CREATE TABLE `drivers` (
```
andrii-demchenko, line 2: Also not sure if it is ok to use library_db in this task, but as I can see your previous task was approved with this naming. Let's also ask mentor's opinion

andrii-demchenko (approved):
Commented on some lines, but I am not 100% sure about them, so let's wait for mentor's feedback

src/main/resources/init_db.sql
andrii-tyndiuk: agree, correct!

Resci (approved):

PR #122, opened 2021-10-04
done solution dao-service

Andruwa808:


pustoj0 (approved):
Making one commit "done solution dao-service" is bad practice, but overall it's good

Resci (approved):

PR #127, opened 2021-10-05
Created class Driver and implementations

dandalb:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
+    }
```
Radikhovskiy, line 24: find() methods in Dao layer should return Optional (not in services). In the service layer better call method get() on Optional and return the object.
Цитата з чекліста*

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        return manufacturerDao.get(id);
+    }
```
Radikhovskiy, line 24: Аналогічно

Radikhovskiy (changes requested):

Radikhovskiy (approved):

Resci (approved):

PR #138, opened 2021-11-11
solution has made

ikhominitch:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow(
+                () -> new DataProcessingException("Could not get driver from DAO by id: " + id));
```
MaliukDaria, line 23:
```suggestion
                () -> new DataProcessingException("There is no driver with id: " + id + "in db"));
```

We could get driver by id, but there is no driver with such id in the db

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow(
+                () -> new DataProcessingException("Could not get manufacturer from DAO by id: "
```
MaliukDaria, line 23:
```suggestion
                () -> new DataProcessingException("There is no manufactirer with  id: "
```

MaliukDaria (approved):
Good Job, left some minor comments about exception messages

PR #141, opened 2021-11-29
tested and working

Andrey-Sidorkin:


src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long newId = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
+        String licenceNumber = resultSet.getString("licence_number");
+        Driver driver = new Driver(name, licenceNumber);
+        driver.setId(newId);
+        return driver;
```
KostyaNikiforov, line 115: It is better to rename variable new Id to Id and do not use setters with constructors in one method.
(I had this suggestion in the last hw)

KostyaNikiforov (approved):

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,117 @@
...
+            deleteDriverStatement.setLong(1, id);
+            return deleteDriverStatement.executeUpdate() > 0;
+        } catch (SQLException throwable) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id + " ",
+                    throwable);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long newId = resultSet.getObject("id", Long.class);
```
Sofasmile, line 110: let's rename newId to id

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Driver;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private static DriverDao driverDao;
```
Sofasmile, line 12: should this field be static?

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.List;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public class ManufacturerServiceImpl implements ManufacturerService {
+    @Inject
+    private static ManufacturerDao manufacturerDao;
```
Sofasmile, line 12: same here

Sofasmile (changes requested):

src/main/java/mate/jdbc/service/DriverServiceImpl.java
Andrey-Sidorkin: why not? why should we create a new object of DriverServiceImpl every time we need it's method?

Sofasmile (approved):

PR #143, opened 2021-11-30
HW Nishchenko

nishchenkokateryna:
…ce layer for Manufacturer


QQBratka (approved):
Im not sure that is OK to not add consructor in Driver like in Manufacturer) but all looks good

Sofasmile (changes requested):
awesome)

Sofasmile (approved):

PR #159, opened 2021-12-02
--> Everything was done according to the terms of the assignment.

SarumanIsAlive:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+    private static final int RANDOM_RANGE = 9999;
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(new Driver("Maria",
+                String.valueOf(new Random().nextInt(RANDOM_RANGE))));
+        driverService.create(new Driver("Alice",
+                String.valueOf(new Random().nextInt(RANDOM_RANGE))));
```
SonyaPakhomova, line 20: are you sure about using random?

SonyaPakhomova (changes requested):

src/main/java/mate/jdbc/Main.java
SarumanIsAlive: согласен, могут выпасть одинаковые значения лицензий

SonyaPakhomova (approved):

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,16 @@
-CREATE SCHEMA IF NOT EXISTS `taxi` DEFAULT CHARACTER SET utf8;
-USE `taxi`;
+CREATE DATABASE `taxi_db` DEFAULT CHARACTER SET utf8 DEFAULT ENCRYPTION='N';
```
Sofasmile, line 2: let's add `USE taxi_db ; `query

Sofasmile (approved):

PR #160, opened 2021-12-02
Hw13

Eugene-exe:


nishchenkokateryna (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,72 @@
...
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerDao manufacturerDao = (ManufacturerDao)
```
Sofasmile, line 13: we should use service here instead of dao

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,17 @@
-CREATE SCHEMA IF NOT EXISTS `taxi` DEFAULT CHARACTER SET utf8;
-USE `taxi`;
+CREATE DATABASE `taxi_service_db` DEFAULT CHARACTER SET utf8;
```
Sofasmile, line 2: let's add query: USE `taxi_service_db `; after db creating

Sofasmile (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,17 +1,17 @@
...
 import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
 
 public class Main {
     private static Injector injector = Injector.getInstance("mate.jdbc");
 
     public static void main(String[] args) {
-        ManufacturerDao manufacturerDao = (ManufacturerDao)
-                injector.getInstance(ManufacturerDao.class);
+        ManufacturerService manufacturerDao = (ManufacturerService)
```
Sofasmile, line 13:
```suggestion
        ManufacturerService manufacturerService = (ManufacturerService)
```

Sofasmile (approved):

PR #181, opened 2021-12-23
implemented task dao-service

alex-hazniuk:


src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+                        = connection.prepareStatement(query)) {
+            List<Driver> drivers = new ArrayList<>();
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                drivers.add(getDriver(resultSet));
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of drivers "
+                    + "from drivers table.", e);
```
MaliukDaria, line 70:
```suggestion
            throw new DataProcessingException("Couldn't get all drivers ", e);

```

we could have different lists of drivers selected by a specific parameter in the future

MaliukDaria (approved):

PR #183, opened 2022-01-17
added model Driver, DriverDao, DriverService

ev-lanovenko:


src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+                 PreparedStatement deleteDriverStatement =
+                         connection.prepareStatement(deleteRequest)) {
+            deleteDriverStatement.setLong(1, id);
+            return deleteDriverStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't delete a driver by id" + id, e);
+        }
+    }
+
+    public Driver getDriver(ResultSet resultSet) {
```
MaliukDaria, line 101: do we need this method to be public

src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Driver get(Long id);
```

Usually we return Optional from DAO layer and Instance from Service

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Manufacturer get(Long id);
```

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,17 @@
...
+  `name` varchar(255) NOT NULL,
+  `country` varchar(255) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=8 DEFAULT CHARSET=utf8mb3;
+
+CREATE TABLE `drivers` (
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `licenseNumber` varchar(255) NOT NULL,
```
MaliukDaria, line 14:
```suggestion
  `license_number` varchar(255) NOT NULL,
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #186, opened 2022-01-31
The class Driver was created.

Viktor-Stadnik:
Added service classes

### DIFF
--- a/src/main/java/mate/jdbc/service/DriverService.java
+++ b/src/main/java/mate/jdbc/service/DriverService.java
@@ -1,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
+
+    List<Driver> getAll();
+
+    Driver update(Driver driver);
+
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
```
MaliukDaria, line 11:
```suggestion
    Driver get(Long id);
```

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
MaliukDaria, line 11:
```suggestion
    Manufacturer get(Long id);
```

MaliukDaria (changes requested):

PR #196, opened 2022-02-17
Jv dao service solution

liliia-ponomarenko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver();
+        driver.setName("Lili");
+        driver.setLicenseNumber("mn2345");
+        driverService.create(driver);
+        driverService.update(new Driver(6L, "Liliya", "nn8900"));
+        driverService.delete(5L);
```
Sasha-Bachynskyi, line 19: who is the 5th driver?
are you sure if I clone your project, this code will work correctly?

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+        driverService.getAll().forEach(System.out::println);
+        System.out.println(driverService.get(3L));
```
Sasha-Bachynskyi, line 21: the same here

Sasha-Bachynskyi (approved):
take a look at my comments)
nice work 🔥

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,26 @@
...
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver();
+        driver.setName("Lili");
+        driver.setLicenseNumber("mn2345");
+        driverService.create(driver);
+        driverService.update(new Driver(6L, "Liliya", "nn8900"));
```
KostyaNikiforov, line 18: Do we need to set `id` while creating a new driver, isn't it AUTO_INCREMENT

src/main/java/mate/jdbc/Main.java
KostyaNikiforov: I agree it's better to use id of drivers you've created

KostyaNikiforov (approved):

PR #197, opened 2022-02-19
hw-jv-dao-service

yurii-tomashevskyi:


pikador10 (approved):
Look great to me  ! !

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,24 @@
...
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver(4L,"Yuriy", "sm002244");
```
KostyaNikiforov, line 14: Do we need to set id here, isn't it AUTO_INCREMENT?

KostyaNikiforov (approved):
Great job!

src/main/java/mate/jdbc/Main.java
yurii-tomashevskyi: In README class Driver already was with field id. That`s why I create constructor with id. But in my table drivers in BD I use AUTO_INCREMENT

PR #200, opened 2022-03-01
Driver model was created and added to DB along with its service and Dao.

karma-o:


src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,19 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+
+@Dao
```
MaliukDaria, line 8: we need to place @Dao annotation over class implementation of interface

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        ) {
+            List<Driver> drivers = new ArrayList<>();
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                drivers.add(getDriver(resultSet));
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of drivers "
+                    + "from drivers table.", e);
```
MaliukDaria, line 75:
```suggestion
            throw new DataProcessingException("Couldn't get all drivers ", e);
```

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -2,8 +2,10 @@
 
 import java.util.List;
 import java.util.Optional;
+import mate.jdbc.lib.Dao;
 import mate.jdbc.model.Manufacturer;
 
+@Dao
```
MaliukDaria, line 8: same here, we don't need to place this annotation here

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public interface ManufacturerService {
```
MaliukDaria, line 8:
```suggestion

public interface ManufacturerService {
```

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #202, opened 2022-03-09
Develop

OleksiyTitov:


src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Driver get(Long id);
```

usually we return Optional from dao layer and entity from service layer

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Manufacturer get(Long id);
```

same here

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #214, opened 2022-03-21
Task done. Codestyle passed.

lugburz-hr:


black-bit0 (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
-        // test your code here
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Audi");
+        manufacturer.setCountry("Germany");
+        manufacturerService.create(manufacturer);
+        manufacturer.setName("Peugeot");
+        manufacturer.setCountry("France");
+        manufacturer = manufacturerService.create(manufacturer);
```
boroda4436, line 22: You are creating the same manufacturer twice? This is not a good practice. Better, create two separate variables, for example, `audi` and `peugeot`, and save them.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        manufacturer.setName("Renault");
+        manufacturerService.update(manufacturer);
+        List<Manufacturer> allManufacturers = manufacturerService.getAll();
+        allManufacturers.stream().forEach(System.out::println);
```
boroda4436, line 26:
```suggestion
        allManufacturers.forEach(System.out::println);
```
stream is redundant

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver();
+        driver.setName("SomeName");
+        driver.setLicenseNumber("12345");
+        driverService.create(driver);
+        driver.setName("NameSome");
+        driver.setLicenseNumber("67890");
+        driver = driverService.create(driver);
```
boroda4436, line 35: same as with manufacturers

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        driver.setName("42");
+        driverService.update(driver);
+        List<Driver> allDrivers = driverService.getAll();
+        allDrivers.stream().forEach(System.out::println);
```
boroda4436, line 39: same

boroda4436 (changes requested):

src/main/java/mate/jdbc/Main.java
boroda4436: А тут?

boroda4436 (changes requested):

boroda4436 (approved):

PR #219, opened 2022-03-21
BUILD success

spartiv:
implemented all methods

### DIFF
--- a/src/main/java/mate/jdbc/model/Driver.java
+++ b/src/main/java/mate/jdbc/model/Driver.java
@@ -46,15 +46,9 @@ public class Driver {
     @Override
     public String toString() {
         return "Driver{"
-                +
-                "id=" + id
-                +
-                ", name='"
-                + name + '\''
-                +
-                ", licenseNumber='"
-                + licenseNumber + '\''
-                +
-                '}';
+                + "id = " + id
+                + ", name = '" + name + '\''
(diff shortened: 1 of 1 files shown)


OleksiyTitov (changes requested):

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
OleksiyTitov, line 22: What kind of exception do you throw?

OleksiyTitov, line 22:
```suggestion
        return manufacturerDao.get(id).orElseThrow(
            () -> new NoSuchElementException("Couldn't get manufacturer by id: " + id)
        );
```

OleksiyTitov (changes requested):

OleksiyTitov (approved):

boroda4436 (changes requested):

boroda4436 (approved):

PR #223, opened 2022-04-05
Implemented second step

tanya-logatskaya:


src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private Driver getDriver(ResultSet resultSet) {
+        Driver driver = new Driver();
+        try {
+            driver.setId(resultSet.getObject("id", Long.class));
+            driver.setName(resultSet.getString("name"));
+            driver.setLicenseNumber(resultSet.getString("licenseNumber"));
+        } catch (SQLException e) {
+            e.printStackTrace();
```
Sofasmile, line 110: let's throw an exception here instead of `e.printStackTrace();`

Sofasmile (changes requested):

Sofasmile (approved):

PR #231, opened 2022-04-29
added ManufacturerService class and DriverService with methods create…

Alina-Vasylyshyn:
…, get, getAll, update, delete

### DIFF
--- a/src/main/java/mate/jdbc/model/Car.java
+++ b/src/main/java/mate/jdbc/model/Car.java
@@ -1,0 +1,10 @@
+package mate.jdbc.model;
+
+import java.util.List;
+
+public class Car {
+    private Long id;
+    private String model;
+    private Manufacturer manufacturer;
+    private List<Driver> drivers;
+}
--- a/src/main/java/mate/jdbc/model/Driver.java
+++ b/src/main/java/mate/jdbc/model/Driver.java
@@ -2,20 +2,20 @@ package mate.jdbc.model;
 
 import java.util.Objects;
 
-public class Manufacturer {
+public class Driver {
     private Long id;
     private String name;
-    private String country;
+    private Long licenseNumber;
 
-    public Manufacturer(Long id, String name, String country) {
+    public Driver(Long id, String name, Long licenseNumber) {
         this.id = id;
         this.name = name;
-        this.country = country;
+        this.licenseNumber = licenseNumber;
     }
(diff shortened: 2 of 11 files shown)


src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,10 @@
+package mate.jdbc.model;
+
+import java.util.List;
+
+public class Car {
```
MaliukDaria, line 5: Why do we need this class is this task?

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class Driver {
+    private Long id;
+    private String name;
+    private Long licenseNumber;
+
+    public Driver(Long id, String name, Long licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
```
MaliukDaria, line 13:
```suggestion
        this(name, licenseNumber);
        this.id = id;
```

MaliukDaria (changes requested):
What about init_db.sql file?

src/main/java/mate/jdbc/model/Car.java
Alina-Vasylyshyn: I agree, dont need

src/main/java/mate/jdbc/model/Driver.java
Alina-Vasylyshyn: ok

src/main/resources/init_db.sql
```diff
@@ -0,0 +1,15 @@
...
+    'licenseNumber' VARCHAR(255) NULL ,
+    'is_deleted' TINYINT NOT NULL  DEFAULT  0,
+    PRIMARY KEY ('id')
+);
+
+CREATE TABLE 'manufacturers' (
+    'id' INT NOT NULL AUTO_INCREMENT,
+    'name' VARCHAR(255) NULL,
+    'country' VARCHAR(255) NULL,
+    'is_deleted' TINYINT NOT NULL  DEFAULT  0
```
MaliukDaria, line 14: Does this script works? What about PRIMARY KEY?

MaliukDaria (approved):

PR #232, opened 2022-05-01
Jv dao service solution

bigmozes:
Created classes and implemented interfaces, application's work tested in Main method, as mentioned in README file.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,37 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Injector injector = Injector.getInstance("mate.jdbc");
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver driver1 = new Driver("Bob", "123456");
+        Driver driver2 = new Driver("Alice", "654321");
+        Driver driver3 = new Driver("John", "123654");
+        driverService.create(driver2);
+        driverService.create(driver3);
+        driverService.create(driver1);
+        driverService.update(new Driver(1L, "Patricia", "918273"));
+        driverService.delete(2L);
+        System.out.println(driverService.get(1L));
+        driverService.getAll().forEach(System.out::println);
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,37 @@
...
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Injector injector = Injector.getInstance("mate.jdbc");
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver driver1 = new Driver("Bob", "123456");
+        Driver driver2 = new Driver("Alice", "654321");
+        Driver driver3 = new Driver("John", "123654");
```
MaliukDaria, line 16: It's better not to use numbers in variables names

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,37 @@
...
+        driverService.create(driver2);
+        driverService.create(driver3);
+        driverService.create(driver1);
+        driverService.update(new Driver(1L, "Patricia", "918273"));
+        driverService.delete(2L);
```
MaliukDaria, line 21:
```suggestion
        driverService.delete(driver2.getId());
```

and so on

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,37 @@
...
+        Injector injector = Injector.getInstance("mate.jdbc");
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver driver1 = new Driver("Bob", "123456");
+        Driver driver2 = new Driver("Alice", "654321");
+        Driver driver3 = new Driver("John", "123654");
+        driverService.create(driver2);
+        driverService.create(driver3);
+        driverService.create(driver1);
+        driverService.update(new Driver(1L, "Patricia", "918273"));
```
MaliukDaria, line 20:
```suggestion
        driverService.update(new Driver(driver1.getId(), "Patricia", "918273"));
```

MaliukDaria (changes requested):

src/main/java/mate/jdbc/Main.java
bigmozes: Thought since it's for the testing purposes it would be a "state of nonsense" to care about.
Refactored code anyway, added using method Create()'s returned value.
Thanks!

MaliukDaria: Ideally, our code should be tested using unit tests, and there, we should follow the code style, so let's stick to it in Main class too 🙂

PR #233, opened 2022-05-01
classes created

oleg-new:
ManufacturerServiceImpl and DriverServiceImpl
the call from Main goes to the "Service" layer and not to the "Dao" layers

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,50 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,50 @@
...
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Injector injector = Injector.getInstance("mate.jdbc");
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        Manufacturer fiat = new Manufacturer(null, "fiat", "Italy");
```
MaliukDaria, line 15: lets use cinstructor with no id as incoming parameter, not to pass nuls

MaliukDaria (changes requested):

MaliukDaria (changes requested):
I dont see changes from previous review

PR #237, opened 2022-05-06
Task Dao service

MarkShapka:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,58 @@
...
+        driverService.create(driverBen);
+        driverService.create(driverBob);
+
+        List<Driver> allDrivers = driverService.getAll();
+        allDrivers.stream()
+                .forEach(System.out::println);
+
+        System.out.println(driverService.get(driverBen.getId()));
+
+        Driver updateDriverBen = new Driver(1L, "Ben", "5656565");
```
MaliukDaria, line 31:
```suggestion
        Driver updateDriverBen = new Driver(driverBen.getId(), "Ben", "5656565");
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement getDriverFromDB = connection.prepareStatement(query)) {
+            getDriverFromDB.setLong(1, id);
+            ResultSet resultSet = getDriverFromDB.executeQuery();
+            if (resultSet.next()) {
+                return Optional.of(parseDriver(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't find driver into DB where id = "
+                    + id, e);
```
MaliukDaria, line 51:
```suggestion
            throw new DataProcessingException("Can't get driver with id = "
                    + id, e);
```

We get this exception while smth went wrong while we getting driver from db, but it could exist in db

MaliukDaria (approved):

PR #265, opened 2022-05-14
jv-dao-service, could you check it, please

lutskvadym:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,44 @@
...
+        manufacturerService.create(new Manufacturer(1L, "Tesla", "USA"));
+        manufacturerService.create(new Manufacturer(1L, "Toyota", "Japan"));
+        System.out.println(System.lineSeparator());// get all
+        List<Manufacturer> allManufacturers = manufacturerService.getAll();
+        allManufacturers.stream()
+                .forEach(System.out::println);
+        //delete & update
+        manufacturerService.delete(2L);
+        manufacturerService.update(new Manufacturer(1L,
+                "Infiniti", "Japan"));
```
MaliukDaria, line 39:
```suggestion
        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
        Driver alice = driverService.create(new Driver("Alice", "11111178"));
        Driver tom = driverService.create(new Driver("Tom", "00000078"));
        Driver anna = driverService.create(new Driver("Anna", "20202020"));
        List<Driver> allDrivers = driverService.getAll();//print all drivers
        allDrivers
                .forEach(System.out::println);
        System.out.println(System.lineSeparator());//update & delete driver
        driverService.update(new Driver(alice.getId(), "Maria", "7777777"));
        driverService.delete(tom.getId());
        driverService.getAll().stream().forEach(System.out::println);
```

same for manufacturers

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        Optional<Driver> driver = driverDao.get(id);
+        return driver.orElse(null);
```
MaliukDaria, line 24: It's better to throw some exception in case there is no driver with such id

MaliukDaria: whta about this comment?

MaliukDaria (changes requested):
not all comments are fixed

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException("Could not get driver from DAO "
+                        + "by id = " + id));
+    }
```
MaliukDaria, line 26:
```suggestion
        return driverDao.get(id)
                .orElseThrow(() -> new NoSuchElementException("There is no driver  "
                        + "with id = " + id));
    }
```

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException("Could not get manufacturer "
+                        + "from DAO by id = " + id));
```
MaliukDaria, line 25:
```suggestion
                .orElseThrow(() -> new NoSuchElementException("There is no manufacturer "
                        + "with id = " + id));
```

MaliukDaria (approved):

PR #272, opened 2022-06-08
implemented service layer

viktoriia-haraniuk:
- Created Driver class
- Created ManufacturerService and implemented required methods
- Created DriverDao and DriverService and implemented required methods
- Tested functionality in Main

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -33,7 +33,7 @@ public class Main {
         driver.setLicenseNumber("XZ642387");
         driverService.update(driver);
 
-        driverService.delete(1L);
+        driverService.delete(driver.getId());
 
         System.out.println(driverService.getAll());
 
--- a/src/main/java/mate/jdbc/service/DriverServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/DriverServiceImpl.java
@@ -18,7 +18,8 @@ public class DriverServiceImpl implements DriverService {
 
     @Override
     public Driver get(Long id) {
-        return driverDao.get(id).orElseThrow();
+        return driverDao.get(id)
+                .orElseThrow(() -> new RuntimeException("Can't get driver. ID:" + id));
     }
 
     @Override
--- a/src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
@@ -18,7 +18,8 @@ public class ManufacturerServiceImpl implements ManufacturerService {
 
     @Override
     public Manufacturer get(Long id) {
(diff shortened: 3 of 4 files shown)


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 21:
```suggestion
        return driverDao.get(id).orElseThrow();
```
let's create a throw with an informative  message

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 21: the same here

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,17 @@
-CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
-USE `taxi_service`;
+CREATE SCHEMA IF NOT EXISTS taxi_service DEFAULT CHARACTER SET utf8;
+USE taxi_service;
```
liliia-ponomarenko, line 2:
```suggestion
CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
USE `taxi_service`;
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,69 @@
...
+
+        driver.setName("Sergiy");
+        driver.setLicenseNumber("ER64232");
+        driverService.create(driver);
+
+        driver.setName("Alina");
+        driver.setName("SD078932");
+        driverService.create(driver);
+
+        System.out.println(driverService.get(1L));
```
liliia-ponomarenko, line 28: this code always will work correctly?
better use `driverService.get(driver.getId())`

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it 😉

liliia-ponomarenko (approved):
good job 👍

PR #276, opened 2022-06-10
1th try: 1) Created Driver Class 2)Created ManufacturerService 3) Cre…

ihor-dmytrush:
…ated DriverDao and DriverService

### DIFF
--- a/src/main/java/mate/jdbc/service/DriverServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/DriverServiceImpl.java
@@ -1,0 +1,38 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Driver;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElse(new Manufacturer());
```
Nick97-git, line 21: in case if manufacturer isn't found it's better to throw an exception

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElse(new Driver());
```
Nick97-git, line 21: in case if manufacturer isn't found it's better to throw an exception

Nick97-git (changes requested):

MaliukDaria (approved):

PR #278, opened 2022-06-14
Dao service

AndrewLinkii:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+    private static int id_Driver = 0;
```
MaliukDaria, line 11:
```suggestion
    private static int driverId = 0;
```

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    private static DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).isPresent() ? driverDao.get(id).get() : null;
```
MaliukDaria, line 23:
```suggestion
        return driverDao.get(id).orElseThrow();
```

it's better to throw some exception if there is no driver with such id

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    private static ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).isPresent() ? manufacturerDao.get(id).get() : null;
```
MaliukDaria, line 23: same

MaliukDaria (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        driverService.create(getDefaultDriver());
+        driverService.create(getDefaultDriver());
+
+        driverService.getAll()
+                .forEach(System.out::println);
+
+        driverService.update(new Driver(3L, "defoult3", "333"));
+        driverService.update(new Driver(2L, "defoult2", "2222"));
+
+        driverService.delete(2L);
```
MaliukDaria, line 32:
```suggestion
        Driver first = driverService.create(getDefaultDriver());
        Driver secont = driverService.create(getDefaultDriver());
        Driver third = driverService.create(getDefaultDriver());

        driverService.getAll()
                .forEach(System.out::println);

        driverService.update(new Driver(third.getId(), "defoult3", "333"));
        driverService.update(new Driver(secont.getId(), "defoult2", "2222"));

        driverService.delete(secont.getId());
```

It's better not to use magic numbers

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #281, opened 2022-06-17
Implement task

gdpbalt:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow(()
+                -> new RuntimeException("Couldn't get driver by id " + id));
+    }
```
Sm0kk, line 24: image

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow(()
+                -> new RuntimeException("Couldn't get manufacturer by id " + id));
```
Sm0kk, line 23: Same

Sm0kk (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
gdpbalt: I disagree with that point.
If we change orElseThrow() to get() we'll get NPE in the fiture

Sm0kk (approved):

liliia-ponomarenko (approved):
good job 👍

PR #293, opened 2022-06-30
Hw jv dao service

Oleksii-Mospan:


src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Driver;
+
+public interface DriverDao {
+
+    Driver create(Driver driver);
+
+    Driver get(Long id);
```
MaliukDaria, line 10:
```suggestion
    Optional<Driver> get(Long id);
```

It's better to return Optional from Dao layer

src/main/java/mate/jdbc/util/Injector.java
```diff
@@ -10,6 +10,9 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
 
 public class Injector {
```
MaliukDaria, line 17: why do we need to move Injector from lib folder?

MaliukDaria (changes requested):

PR #295, opened 2022-07-03
First commit. All needed classes and interfaces were created. Application is working, and CheckStyle runs clear.

MKaramavrov:


src/main/resources/init_db.sql
```diff
@@ -7,3 +7,9 @@ CREATE TABLE `manufacturers` (
                                         `country` VARCHAR(225) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+CREATE TABLE `taxi_service`.`drivers` (
+                                          `id` BIGINT(11) NOT NULL,
```
zim10093, line 11:
```suggestion
                                          `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
```

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        return manufacturerDao.get(id);
+    }
```
zim10093, line 24: On the service layer, you should return the right object(Manufacturer) or throw an exception

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
+    }
```
zim10093, line 24: On the service layer, you should return right object(Driver) of throw exception

zim10093 (changes requested):

zim10093 (approved):

src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -4,19 +4,27 @@
...
         } catch (ClassNotFoundException e) {
             throw new RuntimeException("Can't find SQL Driver", e);
         }
+        Scanner scanLogin = new Scanner(System.in);
+        System.out.println("Enter your mySQL Login");
+        USERNAME = scanLogin.nextLine();
+
+        Scanner scanPassword = new Scanner(System.in);
+        System.out.println("Enter your mySQL Password");
+        PASSWORD = scanPassword.nextLine();
```
dSkarzh, line 27: do it using only one scanner

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,54 @@
...
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
dSkarzh, line 20: try to avoid situation like this with large numbers of constructors, use setters instead

dSkarzh (changes requested):

dSkarzh (approved):

PR #314, opened 2022-07-26
Solution

losnazar:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        Optional<Driver> driverOptional = driverDao.get(id);
+        return driverOptional.get();
```
MaliukDaria, line 24:
```suggestion
        return driverDao.get(id).get();
```

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        Optional<Manufacturer> manufacturerOptional = manufacturerDao.get(id);
+        return manufacturerOptional.get();
```
MaliukDaria, line 24:
```suggestion
        return manufacturerDao.get(id).get();
```

MaliukDaria (approved):

MaliukDaria:
You should not close the PR`s after review, just leave them open

PR #320, opened 2022-08-10
hw-jv-dao-service

serhii-davydov:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+        Driver adam = new Driver("Adam", "AAA-00005");
+        driverService.create(john);
+        driverService.create(kate);
+        driverService.create(pamela);
+        driverService.create(steve);
+        driverService.create(adam);
+        driverService.getAll().forEach(System.out::println);
+
+        System.out.println("Get driver with id: 1.");
+        System.out.println(driverService.get(1L));
```
MaliukDaria, line 26:
```suggestion
        System.out.println(driverService.get(john.getId()));
```

I think It's better not to use "magic numbers" in code 🙂

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+
+        System.out.println("Delete driver with id: 3");
+        System.out.println(driverService.delete(3L));
```
MaliukDaria, line 29:
```suggestion
        System.out.println(driverService.delete(pamela.getId()));
```

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
```
MaliukDaria, line 16:
```suggestion
        this.id = id;
        this(name, licenseNumber);
```

It's better to use one constructor into another

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,11 @@ CREATE TABLE `manufacturers` (
                                         `country` VARCHAR(225) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE drivers
+(
+id SERIAL NOT NULL AUTO_INCREMENT,
+name VARCHAR(30) NOT NULL,
+license_number VARCHAR(30) NOT NULL,
+is_deleted BOOLEAN NOT NULL DEFAULT FALSE
```
MaliukDaria, line 16: I think we should specify primary key here

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,11 @@ CREATE TABLE `manufacturers` (
...
+);
```
MaliukDaria, line 17: dont forget about empty line in the end of file

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #321, opened 2022-08-12
created Driver and implemented the Services and Dao classes

ssplot56:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import mate.jdbc.model.Driver;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    private final DriverDao driverDao;
+
+    public DriverServiceImpl(DriverDao driverDao) {
+        this.driverDao = driverDao;
+    }
+
```
aaananas, line 15:
```suggestion
    @Inject
    private final DriverDao driverDao;
    
```

aaananas (changes requested):

aaananas (approved):

PR #328, opened 2022-08-25
the task is completed

olesyakhalakk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,47 @@
...
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerDao manufacturerDao =
+                (ManufacturerDao) injector.getInstance(ManufacturerDao.class);
+        Manufacturer manufacturerAudi = new Manufacturer(null, "Nissan", "Japan");
```
Rommelua, line 15:
```suggestion
        Manufacturer manufacturerNissan = new Manufacturer(null, "Nissan", "Japan");
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,47 @@
...
+        Manufacturer manufacturerBmw = new Manufacturer(null, "Porsche", "Germany");
```
Rommelua, line 16:
```suggestion
        Manufacturer manufacturerPorsche = new Manufacturer(null, "Porsche", "Germany");
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,47 @@
...
+        manufacturerAudi = manufacturerDao.create(manufacturerAudi);
+        manufacturerBmw = manufacturerDao.create(manufacturerBmw);
```
Rommelua, line 18: See the task: Main should only contain invocation of Service methods, no Dao methods allowed in main.

src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
```
Rommelua, line 10:
```suggestion
    Driver get(Long id);
```

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
Rommelua, line 10:
```suggestion
    Manufacturer get(Long id);
```

Rommelua (changes requested):

src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,16 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import mate.jdbc.model.Driver;
+
+public interface DriverDao {
+    Driver create(Driver driver);
+
+    Driver get(Long id);
```
MaliukDaria, line 9:
```suggestion
    Optional<Driver> get(Long id);
```

we should return Optional from DAO layer, where entity could be null, and the entity itself from Service layer

src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -1,13 +1,12 @@
...
 
 import java.util.List;
-import java.util.Optional;
 import mate.jdbc.model.Manufacturer;
 
 public interface ManufacturerDao {
     Manufacturer create(Manufacturer manufacturer);
 
-    Optional<Manufacturer> get(Long id);
+    Manufacturer get(Long id);
```
MaliukDaria, line 9: same, we should return Optional from Dao layer

src/main/java/mate/jdbc/util/Injector.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+
+public class Injector {
```
MaliukDaria, line 17: why do we need to add second injector here?

MaliukDaria (changes requested):

src/main/java/mate/jdbc/service/ManufacturerService.java
aaananas: up
image

aaananas (changes requested):

aaananas (approved):

PR #332, opened 2022-09-02
mate

Litvik21:


amarkelovUA (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,52 @@
...
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+import mate.jdbc.util.ConnectionUtil;
+
 public class Main {
-    public static void main(String[] args) {
-        // test your code here
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void main(String[] args) throws SQLException {
```
Noct2000, line 14: you need catch these exceptions 
```suggestion
    public static void main(String[] args) {
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,109 @@
...
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of drivers "
+                    + "from drivers table.", e);
+        }
+    }
+
+    @Override
+    public Driver update(Driver driver) {
+        String query = "UPDATE drivers SET name = ?, license_number = ?"
```
Noct2000, line 76: check `is_deleted` column

Noct2000 (changes requested):

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,13 @@ CREATE TABLE `manufacturers` (
...
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
+  `license_number` varchar(255) DEFAULT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=7 DEFAULT CHARSET=utf8mb3;
```
Noct2000, line 17: use consistent case in the `init_db.sql` (upper case or lower case)

Noct2000 (approved):

PR #334, opened 2022-09-03
Hw dao service

MarkGodnyi:


src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private static ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
aaananas, line 21:
```suggestion
        return manufacturerDao.get(id).get();
```
the exception would thrown on dao level

aaananas (approved):

PR #335, opened 2022-09-04
create Driver, create DriverService, ManufacturerService, logic in Main

verbanov:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
amarkelovUA, line 21:
```suggestion
        return driverDao.get(id).get();
```
In this line you call the get method, but in the DAO you return a nullEable. (We use get only notNull values).

amarkelovUA (approved):

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,11 @@ CREATE TABLE `manufacturers` (
...
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
+  `licenceNumber` varchar(255) DEFAULT NULL,
+  `is_deleted` tinyint(1) NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
```
Noct2000, line 17: use consistent case in the `init_db.sql` (upper case or lower case)

Noct2000 (approved):

PR #336, opened 2022-09-05
HW jv-dao-service is done.

Valerii-Sh:


src/main/java/mate/jdbc/dao/ManufacturerDao.java
```diff
@@ -1,11 +1,12 @@
 package mate.jdbc.dao;
 
+import java.sql.SQLException;
 import java.util.List;
 import java.util.Optional;
 import mate.jdbc.model.Manufacturer;
 
 public interface ManufacturerDao {
-    Manufacturer create(Manufacturer manufacturer);
+    Manufacturer create(Manufacturer manufacturer) throws SQLException;
```
MaliukDaria, line 9:
```suggestion
    Manufacturer create(Manufacturer manufacturer);
```

I dont think we need this here

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,74 @@
...
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
MaliukDaria, line 22:
```suggestion
    public Driver(Long id, String name, String licenseNumber) {
        this.id = id;
        this(name, licenseNumber);
    }
```

It is better to use one constructor into another

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -10,6 +10,11 @@ public class Manufacturer {
     public Manufacturer() {
     }
 
+    public Manufacturer(String name, String country) {
```
MaliukDaria, line 13: same here, use one constructor into another

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,15 @@
...
+                                        `name` VARCHAR(255) NOT NULL,
+                                        `country` VARCHAR(255) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+CREATE TABLE `drivers` (
+                                        `id` BIGINT NOT NULL AUTO_INCREMENT,
+                                        `name` varchar(255) NOT NULL,
+                                        `license_number` varchar(255) NOT NULL,
+                                        `is_deleted` TINYINT NOT NULL DEFAULT 0,
+                                        PRIMARY KEY (`id`)) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
MaliukDaria, line 15: dont forget about empty line in the end of file

MaliukDaria (changes requested):

aaananas (approved):

PR #340, opened 2022-09-07
did all the needed tasks and checked common mistakes

GolDenik:


semwithuncle (approved):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Driver driver = new Driver();
+        driver.setName(resultSet.getString(2));
+        driver.setLicenseNumber(resultSet.getString(3));
+        driver.setId(resultSet.getObject(1, Long.class));
```
Sofasmile, line 105: let's use column name instead of indexes

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isPresent()) {
+            return driverDao.get(id).get();
+        }
+        throw new RuntimeException("Couldn't get driver by id " + id);
```
Sofasmile, line 25: it can be replaced with:
```suggestion
        return driverDao.get(id).orElseThrow(() -> 
                new RuntimeException("Couldn't get driver by id " + id);
```

Sofasmile (changes requested):

Sofasmile (approved):

PR #355, opened 2022-09-08
Added new solution.

vadimkik2:


pom.xml
```diff
@@ -22,7 +22,7 @@
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.24</version>
+            <version>8.0.29</version>
```
Nesterenko-Andrey, line 25: you can update to latest version 8.0.30

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,30 @@
...
+        Driver jack = new Driver("Jack","4567");
+        driverService.create(jack);
+
+        driverService.delete(bob.getId());
+        System.out.println(driverService.get(john.getId()));
+
+        john.setLicenseNumber("3456");
+        driverService.update(john);
+        System.out.println(driverService.get(jack.getId()));
+        System.out.println(driverService.getAll());
```
Nesterenko-Andrey, line 28: maybe it would be better to check in main also all method for Manufacturer

Nesterenko-Andrey (changes requested):

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,66 @@
...
+    private String name;
+    private String licenseNumber;
+
+    public Driver(){
+    }
+
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
Noct2000, line 16: If you use this constructor only once, it would be better to remove it, or use it not only once

Noct2000 (changes requested):

Noct2000 (approved):

PR #379, opened 2022-09-09
Jv dao service hw

shevalye:


AshaJaneBrown (approved):
Good job!

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,111 @@
...
+            throw new DataProcessingException("Driver could not be retrieved by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Driver> getAll() {
+        String query = "SELECT * FROM drivers WHERE is_deleted = FALSE";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement
+                        = connection.prepareStatement(query)) {
```
Noct2000, line 61: don't move `=` to the new line, check your code

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,70 @@
...
+    private String licenseNumber;
+
+    public Driver() {
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
Noct2000, line 17: it would be better to remove it, you use `Driver(Long id, String name, String licenseNumber)` only once

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
```
Noct2000, line 7: don't use injectors annotation with interfaces

Noct2000 (changes requested):

Nick97-git (approved):

PR #403, opened 2022-10-28
Add new class and interface (Driver and Manufacturer

IvanAldokhin:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
aaananas, line 22: lets use orElseThrow and throw your eception

aaananas (approved):

PR #405, opened 2022-10-29
added solution

roman680:


src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,17 @@
...
+    Driver create(Driver driver);
+
+    Driver get(Long id);
+
+    List<Driver> getALl();
+
+    Driver update(Driver driver);
+
+    boolean delete(Long id);
+}
```
kateryna-deshchinska, line 17: get() methods in Dao layer should return Optional (not in services). In the service layer better call method get() on Optional and return the object.

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,110 @@
...
+                driver = getDriver(resultSet);
+            }
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get driver by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Driver> getALl() {
```
kateryna-deshchinska, line 56:
```suggestion
    public List<Driver> getAll() {
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,72 @@
...
+        System.out.println("Driver with id = 3 " + driverService.get(3L));
+        System.out.println("Drivers with id = 1 AND 3 were getted successfully"
+                + System.lineSeparator());
+
+        System.out.println("Getting all drivers...");
+        List<Driver> driverList = driverService.getAll();
+        System.out.println("All drivers");
+        for (Driver driver: driverList) {
+            System.out.println(driver);
+        }
```
kateryna-deshchinska, line 58: To display data while testing use Stream API forEach(), not for loop.

kateryna-deshchinska (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isEmpty()) {
+            return driverDao.get(id).get();
+        }
+        return null;
```
kateryna-deshchinska, line 26: If you don't have a driver, better to throw exception
```suggestion
return driverDao.get(id).orElseThrow(<your exeption with message>)
```

src/main/java/mate/jdbc/service/impl/ManufactureServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
kateryna-deshchinska, line 22: same issue

kateryna-deshchinska (changes requested):

PR #414, opened 2022-11-06
HW Dao vs Service

CyberDogFK:


src/main/java/mate/jdbc/services/impl/ManufacturerService.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        return manufacturerDao.get(id);
+    }
```
arthur4655, line 23: Im not sure that we could return empty object

src/main/java/mate/jdbc/services/impl/DriverService.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
+    }
```
arthur4655, line 23: Maybe it is better to return something else but not empty object? But Im not sure about it

arthur4655 (approved):

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,18 @@ CREATE TABLE `manufacturers` (
...
+                                        `is_deleted` TINYINT NOT NULL DEFAULT 0,
+                                        PRIMARY KEY (`id`));
+
+-- Just for test
+
+INSERT INTO drivers (name, license_number)
+    -> VALUES
+    -> ("Anton", "457434"),
+    -> ("Kate", "5646453"),
+    -> ("Vlad", "5436356");
```
liliia-ponomarenko, line 24: Let’s  use methods for testing inserting data

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                 PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
```
liliia-ponomarenko, line 106: If you use try-catch you can show that an exception occurred when receiving data from ResultSet.

src/main/java/mate/jdbc/services/impl/DriverService.java
```diff
@@ -0,0 +1,39 @@
+package mate.jdbc.services.impl;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.model.Driver;
+import mate.jdbc.services.Service;
+
+@mate.jdbc.lib.Service
```
liliia-ponomarenko, line 10:
```suggestion
@Service

```

src/main/java/mate/jdbc/services/impl/ManufacturerService.java
```diff
@@ -0,0 +1,39 @@
+package mate.jdbc.services.impl;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.services.Service;
+
+@mate.jdbc.lib.Service
```
liliia-ponomarenko, line 10: the same here

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,29 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.services.Service;
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Injector injector = Injector.getInstance("mate.jdbc");
```
liliia-ponomarenko, line 9: your injector can be constant

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solutione😉

src/main/java/mate/jdbc/services/impl/ManufacturerService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow(() ->
+                new RuntimeException("No object in DB with id=" + id));
```
liliia-ponomarenko, line 23:
```suggestion
                new DataNotFoundExeption("No object in DB with id=" + id));
```
better use custom exception here

liliia-ponomarenko (approved):
good job! 🔥

PR #419, opened 2022-11-08
Developed one more part of taxi service assignment

igorgalenko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,20 @@
...
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driver = new Driver(null, "Dominic", "MA-753159");
+        driverService.create(driver);
+        System.out.println(driverService.get(1L));
+        driver.setLicenseNumber("MA-987654321");
+        driverService.update(driver);
+        driverService.getAll().forEach(System.out::println);
+        driverService.delete(1L);
     }
```
hronograf, line 19: it would be better to also test `ManufacturerService`

hronograf (approved):
Great! 🔥

src/main/java/mate/jdbc/Main.java
igorgalenko: Thanks. ManufacturedService was provided with assignment and I tested it in the previos task)

PR #422, opened 2022-11-10
Created ManufacturerService and its implementation. Created Dao and S…

PavloPolovyi:
…ervice layers for Driver and their implementations. Tested my code in main method.

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,0 +1,113 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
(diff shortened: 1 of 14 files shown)


GHSpeedWagon (approved):
My advices, my pleasure, your choice

src/main/java/mate/jdbc/service/Service.java
```diff
@@ -0,0 +1,15 @@
+package mate.jdbc.service;
+
+import java.util.List;
+
+public interface Service<E, K> {
```
liliia-ponomarenko, line 5: It’s a good solution for the dao layer but not for services

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+                PreparedStatement statement =
+                        connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
```
liliia-ponomarenko, line 106: If you use try-catch you can show that an exception occurred when receiving data from ResultSet.

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 22: You should  throw an exception with an informative message

liliia-ponomarenko (changes requested):
good job! Let’s  fix a few mistakes

liliia-ponomarenko (approved):
Great 😊

PR #423, opened 2022-11-10
solution

vyshnevsky:


src/main/resources/init_db.sql
```diff
@@ -2,8 +2,19 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
...
-                                        `country` VARCHAR(225) NOT NULL,
-                                        `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                                        PRIMARY KEY (`id`));
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `country` varchar(255) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `id_UNIQUE` (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
Veronika-Tretiak, line 11: UNIQUE KEY `id_UNIQUE` (`id`)
);

src/main/resources/init_db.sql
```diff
@@ -2,8 +2,19 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
...
+
+CREATE TABLE `drivers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
+  `licenseNumber` varchar(255) NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `id_UNIQUE` (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
Veronika-Tretiak, line 20: UNIQUE KEY `id_UNIQUE` (`id`)
);

src/main/resources/init_db.sql
```diff
@@ -2,8 +2,19 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
 USE `taxi_service`;
```
Veronika-Tretiak, line 2: Why USE? I used CREATE

Veronika-Tretiak (approved):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        try (PreparedStatement statement
+                     = ConnectionUtil.getConnection().prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
```
liliia-ponomarenko, line 99: If you use try-catch you can show that an exception occurred when receiving data from ResultSet.

src/main/java/mate/jdbc/service/AbstractService.java
```diff
@@ -0,0 +1,15 @@
+package mate.jdbc.service;
+
+import java.util.List;
+
+public interface AbstractService<T> {
```
liliia-ponomarenko, line 5: It’s a good solution for the dao layer but not for services

src/main/resources/init_db.sql
liliia-ponomarenko, line 11:
```suggestion
);
```
You don’t need id_unique because id auto_increment. auto_increment=3 means that id counting  will start  with 3

liliia-ponomarenko (changes requested):
Looks great! Let’s fix a few mistakes!

src/main/java/mate/jdbc/service/AbstractService.java
vyshnevsky: thanks, I got it

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -18,6 +18,10 @@ public Driver(Long id, String name, String licenseNumber) {
         this.licenseNumber = licenseNumber;
     }
 
+    public Driver() {
+
```
KostyaNikiforov, line 22: Empty line

KostyaNikiforov (approved):

PR #429, opened 2022-11-13
Manufacturer service was implemented, Driver model, dao and service was implemented

muzykaVP:


src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -10,6 +10,11 @@ public class Manufacturer {
     public Manufacturer() {
     }
 
+    public Manufacturer(String name, String country) {
+        this.name = name;
+        this.country = country;
+    }
+
```
Bohdan20b, line 17: You don't have to create a new constructor without id.
When you create a new object of Manufacturer just set null value to id. 
That'll do nicely. Again it's just a tip, this method is also fine.

Bohdan20b (approved):
Seems alright. GJ!

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t create new driver. " + driver, e);
+        }
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        Driver driver = null;
```
ChumakAn, line 40: let's create a driver inside of try block

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+            return Optional.ofNullable(driver);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get driver with id = " + id, e);
+        }
+    }
+
+    @Override
+    public List<Driver> getAll() {
+        List<Driver> driver = new ArrayList<>();
```
ChumakAn, line 58: same here, let's create a list of drivers after establishing connection

src/main/resources/init_db.sql
```diff
@@ -2,8 +2,18 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
...
+                                 `is_deleted` TINYINT NOT NULL DEFAULT 0,
+                                 PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+                           `id` bigint NOT NULL AUTO_INCREMENT,
+                           `name` varchar(255) NOT NULL,
+                           `license_number` varchar(255) NOT NULL,
+                           `is_deleted` tinyint NOT NULL DEFAULT '0',
+                           PRIMARY KEY (`id`),
+                           UNIQUE KEY `id_UNIQUE` (`id`),
```
ChumakAn, line 17: id is already unique, cause it's primary key

ChumakAn (changes requested):
Good job, minor comments

ChumakAn (approved):

PR #432, opened 2022-11-14
Implemented Driver model + DriverDao, Driver+ManufacturerService

NSay0:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -81,7 +82,9 @@ public Manufacturer update(Manufacturer manufacturer) {
             statement.setString(1, manufacturer.getName());
             statement.setString(2, manufacturer.getCountry());
             statement.setLong(3, manufacturer.getId());
-            statement.executeUpdate();
+            if (get(manufacturer.getId()).isPresent()) {
+                statement.executeUpdate();
+            }
```
ChumakAn, line 87: it was not mandatory, just an idea

ChumakAn (approved):
Looks good to me

PR #434, opened 2022-11-14
released DriverService and ManufacturerService with their impl

apc10bbe:


anton-pushkar (approved):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,110 @@
...
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update a driver "
+                    + driver, e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE drivers SET is_deleted = TRUE WHERE id = ?";
```
KostyaNikiforov, line 93:
```suggestion
        String query = "UPDATE drivers SET is_deleted = TRUE WHERE id = ? and  is_deleted = FALSE";
```

KostyaNikiforov (approved):
Good job!:fire:

PR #436, opened 2022-11-14
Realized service with dao.

OleksandrKorniychuk:


Mar4uk21 (approved):
LGTM)

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
+    }
```
ChumakAn, line 23: let's use orElseThrow() - and throw NSEE with informational message

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
ChumakAn, line 22: same here

ChumakAn (changes requested):

ChumakAn (approved):

PR #439, opened 2022-11-14
Implemented ManufacturerService, DriverService and DriverDao

denys-domashevskyi:


Yaroslav781 (approved):
Well done!

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow(
+                () -> new RuntimeException("There are no users with such id :" + id));
```
ChumakAn, line 23:
```suggestion
                () -> new RuntimeException("There are no users with such id :" + id));
```
Maybe it's better to throw NoSuchElementException

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow(
+                () -> new RuntimeException("There are no manufacturers with such id :" + id));
```
ChumakAn, line 23: same here

ChumakAn (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
ChumakAn, line 22: Screenshot 2022-11-15 at 11 01 39

it should be with info message too)

ChumakAn (changes requested):

ChumakAn (approved):

PR #440, opened 2022-11-14
created class Driver and manufacturer service and implemented all met...

Yaroslav781:
…hods

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,22 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driverPishta = new Driver("Pishta", "OT103976");
+        Driver driverBob = new Driver("Bob", "ST103942");
+        driverService.create(driverBob);
+        driverService.create(driverPishta);
+        driverBob.setLicenseNumber("PU190953");
+        driverService.update(driverBob);
+        List<Driver> drivers = driverService.getAll();
+        driverService.delete(driverPishta.getId());
     }
 }
--- a/src/main/java/mate/jdbc/service/ManufacturerService.java
+++ b/src/main/java/mate/jdbc/service/ManufacturerService.java
@@ -1,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public interface ManufacturerService {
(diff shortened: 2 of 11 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,22 @@
...
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driverPishta = new Driver("Pishta", "OT103976");
+        Driver driverBob = new Driver("Bob", "ST103942");
+        driverService.create(driverBob);
+        driverService.create(driverPishta);
+        driverBob.setLicenseNumber("PU190953");
+        driverService.update(driverBob);
+        List<Driver> drivers = driverService.getAll();
+        driverService.delete(driverPishta.getId());
```
denys-domashevskyi, line 20: I think we need to test manufacturer too

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        return null;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        return false;
+    }
+}
```
denys-domashevskyi, line 39: I think you forgot to write implementation of these methods

denys-domashevskyi: image

denys-domashevskyi (changes requested):
Nice solution, let's improve it)

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
```
denys-domashevskyi, line 7: Should we use it on interface or just in implementation?

denys-domashevskyi (changes requested):
One more small fix

denys-domashevskyi (approved):
Good job!

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public Manufacturer update(Manufacturer manufacturer) {
+        return null;
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        return false;
```
KostyaNikiforov, line 40: Don't you forget to implement smth?)

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
```
KostyaNikiforov, line 43: Empty line?

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #441, opened 2022-11-14
Hw jv dao service

romangolovatyi:


igorgalenko (approved):
Good job!

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
ChumakAn, line 22: let's throw the more informational exception, NSEE with message

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
ChumakAn, line 22: same here

ChumakAn (changes requested):

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
romangolovatyi: Thanks for your review!

ChumakAn (approved):

PR #442, opened 2022-11-14
solved jv dao service task

HryhorashPavlo:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
```
luka-lozovyk, line 21: This method will throw NoSuchElementException without any message. Pass throw new NSEE("message") for better understanding why exception ocured. Same in ManufacturerServiceImpl

luka-lozovyk (changes requested):

luka-lozovyk (approved):
Seems good!

ChumakAn (approved):
LGTM

PR #452, opened 2022-11-27
jv-dao-service-homework-solution

Pro1OOGamer:
1. the connection parameters have been changed
2. created class driver and its implementation on Dao and Service layers
3. the implementation at the Dao and Service levels for the Manufacturer class has been made
4. @Service and @Dao annotations are inserted in the appropriate places
5. the code is tested in Main
6. entered data to create an SQL table


zachar-chorny (approved):
Good job

PR #453, opened 2022-11-28
Solution was added.

KatyaHumeniuk:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isPresent()) {
+            return driverDao.get(id).get();
+        }
+        throw new RuntimeException("No such id " + id);
```
zachar-chorny, line 24: Lets use orElseThrow instead of get() for it

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        if (manufacturerDao.get(id).isPresent()) {
+            return manufacturerDao.get(id).get();
```
zachar-chorny, line 22: same

zachar-chorny (changes requested):
Good job

zachar-chorny (approved):

PR #469, opened 2023-01-09
added jv-dao-service homework 03 solution

roman-yuskevych:


aleksandr-bevza (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,48 @@
...
+import mate.jdbc.service.ManufacturerService;
+import mate.jdbc.service.ManufacturerServiceImpl;
+
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        System.out.println("Magic STARTS...");
+
+        ManufacturerService manufacturerService = new ManufacturerServiceImpl();
+        DriverService driverService = new DriverServiceImpl();
```
nacenik, line 16: You should get service implementation via Injector class. Fix it in main method.

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.dao.DriverDaoImpl;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Driver;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private static final DriverDao driverDao = new DriverDaoImpl();
```
nacenik, line 13: it should be non-static field. Also, you have annotation `@inject`, you don't need to hardcode implementation 
```suggestion
    private DriverDao driverDao;
```

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+import mate.jdbc.dao.ManufacturerDao;
+import mate.jdbc.dao.ManufacturerDaoImpl;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public class ManufacturerServiceImpl implements ManufacturerService {
+    @Inject
+    private static final ManufacturerDao manufacturerDao = new ManufacturerDaoImpl();
```
nacenik, line 13: The same as in DriverServiceImpl

nacenik (changes requested):

aaananas (approved):

PR #470, opened 2023-01-09
Created ManufacturerService and DriverService

SofiiaKazanivska:


src/main/java/mate/jdbc/services/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
aaananas, line 11: image
what return type should it be?

aaananas (changes requested):

aaananas (approved):

PR #476, opened 2023-01-16
Created class Driver

metelska-anastasiia:
Created ManufacturerService and its implementation, added annotations where needed Created Dao and Service layers for Driver and their implementations Added create statement for drivers table to sql file Changed constants in ConnectionUtil class
Checked methods in Main

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,36 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        //Driver driver = new Driver();
+        //driver.setName("Ann");
+        //driver.setLicenseNumber("5555");
+        //driver.setId(5L);
+        //driverService.create(driver);
+        //System.out.println(driverService.getAll());
+        //System.out.println(driverService.get(5L));
+        //driverService.delete(5L);
+        //driverService.update(driver);
+        //System.out.println(driverService.getAll());
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Renault");
+        manufacturer.setCountry("France");
+        manufacturer.setId(13L);
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        //manufacturerService.create(manufacturer);
+        //manufacturerService.delete(5L);
+        //manufacturerService.update(manufacturer);
+        System.out.println(manufacturerService.get(16L));
+        System.out.println(manufacturerService.getAll());
+
     }
 }
--- a/src/main/java/mate/jdbc/model/Driver.java
+++ b/src/main/java/mate/jdbc/model/Driver.java
@@ -2,18 +2,18 @@ package mate.jdbc.model;
 
 import java.util.Objects;
 
-public class Manufacturer {
+public class Driver {
     private Long id;
     private String name;
-    private String country;
+    private String licenseNumber;
 
-    public Manufacturer() {
+    public Driver() {
     }
 
-    public Manufacturer(Long id, String name, String country) {
+    public Driver(Long id, String name, String licenseNumber) {
         this.id = id;
         this.name = name;
-        this.country = country;
+        this.licenseNumber = licenseNumber;
     }
 
     public Long getId() {
@@ -32,12 +32,12 @@ public class Manufacturer {
         this.name = name;
     }
 
-    public String getCountry() {
-        return country;
+    public String getLicenseNumber() {
+        return licenseNumber;
     }
 
-    public void setCountry(String country) {
-        this.country = country;
+    public void setLicenseNumber(String licenseNumber) {
+        this.licenseNumber = licenseNumber;
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
+        //driver.setName("Ann");
+        //driver.setLicenseNumber("5555");
+        //driver.setId(5L);
+        //driverService.create(driver);
+        //System.out.println(driverService.getAll());
+        //System.out.println(driverService.get(5L));
+        //driverService.delete(5L);
+        //driverService.update(driver);
+        //System.out.println(driverService.getAll());
+        Manufacturer manufacturer = new Manufacturer();
```
okuzan, line 23:
```suggestion
        Manufacturer renault = new Manufacturer();
```

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, licenseNumber);
+    }
+
+    @Override
+    public String toString() {
+        return System.lineSeparator() + "Driver{"
```
okuzan, line 64: better use lineSeparator at the end

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        Driver olga = new Driver();
+        olga.setName("Olga");
+        olga.setLicenseNumber("5678");
+        olga.setId(10L);
```
nacenik, line 16: you don't need to set the id for the driver. This id will be set up automatically after creating row in DB

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(olga);
+        System.out.println("Get all drivers from drivers table"
+                + driverService.getAll());
+        System.out.println("Get driver with id = 5 from drivers table "
```
nacenik, line 21: why id = 5? if you get id = 1?

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+                + driverService.get(1L));
+        System.out.println("Deleting driver with id = 5 from drivers table"
+                + driverService.delete(5L));
```
nacenik, line 24: Do you have driver with id 5?

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        Driver olga = new Driver();
+        olga.setName("Olga");
+        olga.setLicenseNumber("5678");
+        olga.setId(10L);
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(olga);
+        System.out.println("Get all drivers from drivers table"
+                + driverService.getAll());
+        System.out.println("Get driver with id = 5 from drivers table "
+                + driverService.get(1L));
```
nacenik, line 22: better don't use the magic number, and do smth like this: `olga.getId()`

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        driverService.update(olga);
+        driverService.getAll().forEach(System.out::println);
+
+        Manufacturer ford = new Manufacturer();
+        ford.setName("Ford");
+        ford.setCountry("China");
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        manufacturerService.create(ford);
+        manufacturerService.delete(12L);
```
nacenik, line 35: do you have this id in DB?

nacenik (changes requested):

src/main/java/mate/jdbc/Main.java
metelska-anastasiia: because in my table after tests I have a lot of drivers))

src/main/java/mate/jdbc/Main.java
metelska-anastasiia: I have driver even with id 12)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+
+        Manufacturer honda = new Manufacturer();
+        honda.setName("Honda");
+        honda.setCountry("China");
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        manufacturerService.create(honda);
+        honda.setCountry("Japan");
+        manufacturerService.update(honda);
+        System.out.println("Getting manufacturer Ford "
```
okuzan, line 36:
```suggestion
        System.out.println("Getting manufacturer Honda "
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
+        mykyta.setLicenseNumber("1111");
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(mykyta);
+        System.out.println("Get all drivers from drivers table"
+                + driverService.getAll());
+        System.out.println("Get driver with id = 1 from drivers table "
+                + driverService.get(mykyta.getId()));
+        mykyta.setLicenseNumber("2222");
+        driverService.update(mykyta);
+        System.out.println("Deleting driver with id = 5 from drivers table"
```
okuzan, line 24:
```suggestion
        System.out.println("Deleting Mykyta from drivers table"
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,41 @@
...
     public static void main(String[] args) {
-        // test your code here
+        Driver mykyta = new Driver();
+        mykyta.setName("Mykyta");
+        mykyta.setLicenseNumber("1111");
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(mykyta);
+        System.out.println("Get all drivers from drivers table"
+                + driverService.getAll());
+        System.out.println("Get driver with id = 1 from drivers table "
```
okuzan, line 20:
```suggestion
        System.out.println("Get Mykyta from drivers table "
```

okuzan (approved):
Good job!

PR #477, opened 2023-01-17
created model for a driver, dao for a driver, services for a driver

BusyhinD:
and a manufacturer

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,48 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        //Test operation with driver
+        Driver driver = new Driver();
+        driver.setName("Bob");
+        driver.setLicenseNumber("AB12345");
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driverWithId = driverService.create(driver);
+        driverWithId.setName("Jack");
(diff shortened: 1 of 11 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,48 @@
...
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        //Test operation with driver
+        Driver driver = new Driver();
```
nacenik, line 14: Better to give a more understandable names for variables. Fix it in all places 
```suggestion
        Driver bob = new Driver();
```

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,105 @@
...
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            List<Driver> drivers = new ArrayList<>();
+            ResultSet resultSet = statement.executeQuery();
+            while (resultSet.next()) {
+                drivers.add(getDriver(resultSet));
+            }
+            return drivers;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get list of drivers"
+                    + " from drivers table.", e);
```
nacenik, line 68:
```suggestion
            throw new DataProcessingException("Couldn't get list of drivers from table.", e);
```

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
```
nacenik, line 22: better to add `orElseThrow()` and throw NoSuchElementException with a message. Same for manufacture

nacenik (changes requested):

okuzan (approved):
Good work :)

PR #481, opened 2023-01-18
The classes Driver, DriverDao and DriverService were realized

Roman-Usyk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,13 @@
...
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
```
aaananas, line 10: lets test your code from main, create a few drivers, get them and see what happens

aaananas (approved):

src/main/java/mate/jdbc/Main.java
Roman-Usyk: I tested my code in the main method however i deleted it

PR #482, opened 2023-01-19
HW service VS dao

Maks4u:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
Dmytro-Mk, line 22: lets add message here "Can't get driver by id " + id   or something like this
same for manufacturer service

Dmytro-Mk (changes requested):

Dmytro-Mk (approved):

Nick97-git (approved):

PR #489, opened 2023-01-23
v01

Bohdan-Halitskyi:
Implemented dao for driver table and added service interfaces with im…pls for DriverDao and ManufacturerDao impls, edited init_db.sql to add driver table to scheme, reorganised classes by packages, tested injection (which annotations were added) in Main, which used as driver class, added Car model

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,4 +1,4 @@
-package mate.jdbc.dao;
+package mate.jdbc.dao.impl;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
+        String licenseNumber = resultSet.getString("license_number");
+        return new Driver(id, name, licenseNumber);
```
MaxMaksym, line 110: You don't need variables, you can pass values to the constructor

src/main/java/mate/jdbc/model/Car.java
```diff
@@ -0,0 +1,10 @@
+package mate.jdbc.model;
+
+import java.util.List;
+
+public class Car {
+    private Long id;
+    private String model;
+    private Manufacturer manufacturer;
+    private List<Driver> drivers;
+}
```
MaxMaksym, line 10: We don't need this class

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,51 @@
+package mate.jdbc.model;
+
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver() {
+    }
```
MaxMaksym, line 9: Do you use this constructor?

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
MaxMaksym, line 22: You can throw NSEE in this case and add informative message

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow();
```
MaxMaksym, line 22: same

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,16 @@
...
 
 CREATE TABLE `manufacturers` (
                                         `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
                                         `name` VARCHAR(225) NOT NULL,
                                         `country` VARCHAR(225) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+CREATE TABLE `drivers` (
+                   `id` bigint NOT NULL AUTO_INCREMENT,
+                   `name` varchar(45) NOT NULL,
```
MaxMaksym, line 12:
```suggestion
                   `name` varchar(255) NOT NULL,
```

MaxMaksym (changes requested):

MaxMaksym (approved):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isEmpty()) {
+            throw new NoSuchElementException("Can't get driver with id " + id);
+        }
+        return driverDao.get(id).get();
```
KostyaNikiforov, line 26: What about this option?)
```suggestion
        return driverDao.orElseThrow(() -> new NoSuchElementException("Can't get driver with id " + id));
```

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #490, opened 2023-01-23
Driver task Three-tier architecture implemented 1 commit

ar4ibaldd:


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,107 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            ResultSet resultSet = statement.executeQuery();
+            Driver driver = null;
+            if (resultSet.next()) {
+                return Optional.of(getDriver(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get info about drivers from DB", e);
```
Dmytro-Mk, line 51:
```suggestion
            throw new DataProcessingException("Can`t get driver from DB by id " + id, e);
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,107 @@
...
+        String query = "SELECT * FROM drivers WHERE is_deleted = false";
+        List<Driver> driverList = new ArrayList<>();
+        try (Connection connection = ConnectionUtil.getConnection();
+                Statement statement = connection.createStatement()) {
+            ResultSet resultSet = statement.executeQuery(query);
+            while (resultSet.next()) {
+                driverList.add(getDriver(resultSet));
+            }
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can`t get info about drivers from DB", e);
```
Dmytro-Mk, line 67:
```suggestion
            throw new DataProcessingException("Can`t get list of drivers from DB", e);
```

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
Dmytro-Mk, line 22: lets add message here. ex: "can't get driver by id " + id or something like this
same for manufacturer

Dmytro-Mk (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
ar4ibaldd: I think we don`t need add message here because we throw this massage on DAO tier if we have no driver

Dmytro-Mk (approved):

KostyaNikiforov (approved):

PR #493, opened 2023-01-23
Manufacturer and Driver services created

taras-zhyravetskyi:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
lysytskapolina, line 22: (Suggestion only) Maybe it's better to use orElseThrow method here

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
lysytskapolina, line 22: (Suggestion only) The same about orElseThrow

lysytskapolina (approved):

KostyaNikiforov (approved):

PR #502, opened 2023-01-24
Hw

Mykola-Osolinskyi:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id)
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get driver with id: " + id));
```
KostyaNikiforov, line 24: In such cases it is better to throw NSEE

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id)
+                .orElseThrow(() ->
+                        new RuntimeException("Can't get manufacturer by id: " + id));
```
KostyaNikiforov, line 24: same

KostyaNikiforov (approved):

PR #505, opened 2023-02-01
was solved

Andrii-Varlanovych:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,66 @@
...
+
+        Driver driverWhichWasGot = driverService.get(Long.valueOf(8));
+        System.out.println(driverWhichWasGot);
+        System.out.println("Driver which was got by ID\n");
+
+        driverWhichWasGot.setLicenseNumber("GB_2009");
+        System.out.println(driverService.update(driverWhichWasGot));
+        System.out.println("Driver which was updated\n");
+
+        driverService.delete(Long.valueOf(9));
```
nacenik, line 62: Better ti create constant for magic number 
```suggestion
        driverService.delete(9L);
```

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,55 @@
...
+    public Driver(String name, String licenseNumber) {
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
```
nacenik, line 20: Reuse contractor with 2 parameters
```suggestion
    public Driver(Long id, String name, String licenseNumber) {
        this(name, licenseNumber);
        this.id = id;
    }
```

nacenik (approved):
Good Job!

PR #506, opened 2023-02-02
Added dao/impl and services/impl. Also class driver and table on DB.

oAlExAnDeRo:


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+                         connection.prepareStatement(deleteDriverRequest)) {
+            deleteDriverStatement.setLong(1, id);
+            return deleteDriverStatement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new RuntimeException("Can't delete drivers from DB "
+                    + deleteDriverRequest, e);
+        }
+    }
+
+    private static Driver parseDriver(ResultSet resultSet) {
```
Rommelua, line 112:
```suggestion
    private Driver parseDriver(ResultSet resultSet) {
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        try {
```
Rommelua, line 113: you call this method inside try-catch blocks, so, you do not need try-catch here

Rommelua (changes requested):

Rommelua (approved):

PR #508, opened 2023-02-04
driver and manufacturer services ready

ItsUnbidden:


src/main/resources/init_db.sql
```diff
@@ -1,9 +1,17 @@
...
-                                        `name` VARCHAR(225) NOT NULL,
-                                        `country` VARCHAR(225) NOT NULL,
-                                        `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                                        PRIMARY KEY (`id`));
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `country` varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci NOT NULL,
+  `name` varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci NOT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB AUTO_INCREMENT=36 DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci;
```
karma-o, line 17: Not a mistake, but an overall recommendation, columns do not require an explicit character set, as you did it here.
Also, if you are not changing the functionality too much, it is always recommended to try and stick with the old code arrangement, line breaks, and style, and not move code blocks/lines up and down, when it is not needed.
Doing that can produce more merge conflicts if you are working on the same functionality in parallel with another person.
It is actually an `O` letter from the `SOLID`, Objects or entities should be open for extension but closed for modification

karma-o (approved):
Please check out the comment that I have left

src/main/resources/init_db.sql
ItsUnbidden: I will keep it in mind. Thanks.

PR #514, opened 2023-02-14
All tasks are done.

svitlana-tertyshna:
Created class Driver; created interfaces Dao, Service for models and its implementations.

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,39 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.servise.DriverService;
+
 public class Main {
+    private static final Injector injector =
+            Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        //added driver to DB
+        Driver driverFirst = new Driver("Rita", "B1111");
+        Driver driverSecond = new Driver("Oscar", "C1112");
+        Driver driverThird = new Driver("Moony", "A01");
+        Driver driverForth = new Driver("Galo", "D07");
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(driverFirst);
+        driverService.create(driverSecond);
+        driverService.create(driverThird);
+        Driver createdDriverForth = driverService.create(driverForth);
+        System.out.println("Added data to driver table:");
+        List<Driver> driverList = driverService.getAll();
+        driverList.forEach(System.out::println);
+        //get driver by id and update it
+        System.out.println("Get driver from DB by id = " + createdDriverForth.getId()
+                + " and update info.");
+        createdDriverForth.setLicenseNumber("B007");
+        Driver updateDriverFirst = driverService.update(createdDriverForth);
+        System.out.println(updateDriverFirst);
+        //soft delete driver by id
+        System.out.println("Data from DB after deleted drive by id = "
+                + updateDriverFirst.getId());
+        driverService.delete(updateDriverFirst.getId());
+        driverService.getAll().forEach(System.out::println);
     }
 }
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,4 +1,4 @@
-package mate.jdbc.dao;
+package mate.jdbc.dao.impl;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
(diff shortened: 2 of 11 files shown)


src/main/java/mate/jdbc/servise/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
okuzan, line 22: don't take for granted that you will always have something to `get` :) use `orElseThrow`

src/main/java/mate/jdbc/servise/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
okuzan, line 22: same here

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,39 @@
...
+    private static final Injector injector =
+            Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        //added driver to DB
+        Driver driverFirst = new Driver("Rita", "B1111");
+        Driver driverSecond = new Driver("Oscar", "C1112");
+        Driver driverThird = new Driver("Moony", "A01");
+        Driver driverForth = new Driver("Galo", "D07");
```
okuzan, line 17: bad naming - call them by their real names instead

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,39 @@
...
+        System.out.println("Get driver from DB by id = " + createdDriverForth.getId()
+                + " and update info.");
+        createdDriverForth.setLicenseNumber("B007");
+        Driver updateDriverFirst = driverService.update(createdDriverForth);
+        System.out.println(updateDriverFirst);
+        //soft delete driver by id
+        System.out.println("Data from DB after deleted drive by id = "
+                + updateDriverFirst.getId());
+        driverService.delete(updateDriverFirst.getId());
+        driverService.getAll().forEach(System.out::println);
```
okuzan, line 37: what about manufacturer service?

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,39 @@
...
+        Driver driverFirst = new Driver("Rita", "B1111");
+        Driver driverSecond = new Driver("Oscar", "C1112");
+        Driver driverThird = new Driver("Moony", "A01");
+        Driver driverForth = new Driver("Galo", "D07");
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        driverService.create(driverFirst);
+        driverService.create(driverSecond);
+        driverService.create(driverThird);
+        Driver createdDriverForth = driverService.create(driverForth);
```
okuzan, line 23: just use the same driverFourth, in your `create` method  you return the driver object which you got as an argument, but with `id` set

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,11 @@ CREATE TABLE `manufacturers` (
...
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+                                        `id` int NOT NULL AUTO_INCREMENT,
+                                        `name` varchar(25) NOT NULL,
+                                        `license_number` varchar(255) NOT NULL,
+                                        `is_deleted` tinyint NOT NULL DEFAULT '0',
+                                        PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
```
okuzan, line 17: it's the default so can be omitted

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+                PreparedStatement statement
+                        = connection.prepareStatement(query)) {
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
```
okuzan, line 106: better use try catch with `throw new DataProcessingException`

okuzan (changes requested):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        String name;
+        String licenseNumber;
+        try {
+            id = resultSet.getObject("id", Long.class);
+            name = resultSet.getString("name");
+            licenseNumber = resultSet.getString("license_number");
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get driver from DB", e);
+        }
+        return new Driver(id, name, licenseNumber);
```
aaananas, line 117:
```suggestion
    private Driver getDriver(ResultSet resultSet) throws SQLException {
        Long id = resultSet.getObject("id", Long.class);
        String name = resultSet.getString("name");
        String licenseNumber = resultSet.getString("license_number");
        return new Driver(id, name, licenseNumber);
```

aaananas (approved):

PR #519, opened 2023-02-18
jv-dao-service-solution

nkharyshyn:
-connected to my taxi_service_db Database;
-created service layer for manufacturers;
-created dao and service layers for drivers;
-initialized Dao and Service objects using dependency injection

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,58 @@
 package mate.jdbc;
 
+import java.util.List;
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Long EXISTING_INSTANCE = 1L;
+    private static final Long NOT_EXISTING_INSTANCE = 100L;
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = new Manufacturer(null, "Ford", "USA");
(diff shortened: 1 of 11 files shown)


src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,70 @@
...
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver() {
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
```
Ivan95kos, line 14: We don't have to set id it is automatically generated
```suggestion
    public Driver(String name, String licenseNumber) {
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,58 @@
...
+    private static final Long EXISTING_INSTANCE = 1L;
+    private static final Long NOT_EXISTING_INSTANCE = 100L;
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = new Manufacturer(null, "Ford", "USA");
```
Ivan95kos, line 19: Give more specific names
```suggestion
        Manufacturer ford = new Manufacturer(null, "Ford", "USA");
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,58 @@
...
+        System.out.println(manufacturerService.create(manufacturer));
+        manufacturerService.getAll().forEach(System.out::println);
+
+        Manufacturer updatedManufacturer = manufacturerService.get(EXISTING_INSTANCE);
```
Ivan95kos, line 23: It is better to use the id from the models
```suggestion
        Manufacturer updatedManufacturer = manufacturerService.get(ford.getId());
```

Ivan95kos (changes requested):

kozhukhovsky (approved):
Good job!

PR #528, opened 2023-02-27
Dao service solution

kateryna-mykh:


src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import mate.jdbc.model.Driver;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        Optional<Driver> driverByLicenseNumber =
```
kozhukhovsky, line 17: why do you need this check? for this method it is enough to:
return driverDao.create(driver);

kozhukhovsky (changes requested):

src/main/java/mate/jdbc/service/DriverServiceImpl.java
kateryna-mykh: @ kozhukhovsky, explanation to check in 22-23 line in 'create' method. I think It was logical that the license numbers need to be unique and personal.

kozhukhovsky: @kateryna-mykh That's a good way of thinking about it, but it's not necessary in this particular assignment.

kozhukhovsky (approved):

PR #531, opened 2023-03-03
implemented manufacturer service, added model of driver, implemented …

YevheniiKilovyi:
…driver dao with its implementation, and implemeneted driverservice

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,27 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        final DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver bob = new Driver();
+        bob.setName("Bob");
+        bob.setLicenceNumber("3968722");
+        Driver alice = new Driver();
+        alice.setName("Alice");
+        alice.setLicenceNumber("4472652");
+        Driver john = new Driver();
+        john.setName("John");
+        john.setLicenceNumber("88833162");
+        driverService.create(bob);
+        driverService.create(alice);
+        driverService.create(john);
+        driverService.getAll().forEach(System.out::println);
     }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,27 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
```
okuzan, line 10: what about ManufacturerService?

src/main/java/mate/jdbc/serviceimpl/DriverServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isEmpty()) {
+            throw new NoSuchElementException("Can't get manufacturer with ID " + id);
+        }
+        return driverDao.get(id).get();
```
okuzan, line 27:
```suggestion
        return driverDao.get(id).orElseThrow();

```
throw error explicitly where possible

src/main/java/mate/jdbc/serviceimpl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        if (manufacturerDao.get(id).isEmpty()) {
+            throw new NoSuchElementException("Can't get manufacturer with ID " + id);
+        }
+        return manufacturerDao.get(id).get();
```
okuzan, line 27: same here

okuzan (changes requested):
rename package serviceimpl to just impl

src/main/java/mate/jdbc/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
+package mate.jdbc.impl;
```
kozhukhovsky, line 1:
```suggestion
package mate.jdbc.service.impl;

```

src/main/java/mate/jdbc/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
+package mate.jdbc.impl;
```
kozhukhovsky, line 1: same

kozhukhovsky (approved):

PR #532, opened 2023-03-06
Initial commit

waspp01:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
```
aaananas, line 22: service should not return Optional

aaananas (changes requested):

aaananas (approved):

PR #534, opened 2023-03-09
dao-service solution

DmitriyK86:


uliana-andrushchuk (approved):
Good job! added minor comments

PR #557, opened 2023-03-23
Dao vs Services implemented

copypasterr:


src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -73,7 +74,7 @@ public List<Manufacturer> getAll() {
 
     @Override
     public Manufacturer update(Manufacturer manufacturer) {
-        String query = "UPDATE manufacturers SET name = ?, country = ?"
+        String query = "UPDATE manufacturers SET name = ?, licenseNumber = ?"
```
ArtemPonomar, line 77: Entity Manufacturer doesn't contain property licenseNumber

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+            statement.setLong(1, id);
+            return statement.executeUpdate() > 0;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id " + id, e);
+        }
+    }
+
+    private Driver getDriver(ResultSet resultSet) throws SQLException {
+        Long id = resultSet.getObject("id", Long.class);
+        String name = resultSet.getString("name");
```
ArtemPonomar, line 108: Replace every resultSet.getString({name}); 
with resultSet.getObject({name}, String.class);

ArtemPonomar (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,39 @@
...
+        manufacturerService.update(new Manufacturer("Renault", "France"));
+        List<Driver> drivers = Arrays.asList(
+                new Driver("Petrovych", "DT23453"),
+                new Driver("Ivanovych", "SC25325"),
+                new Driver("Antonovych", "FF35654")
+        );
+        drivers.forEach(driverService::create);
+        Driver driver = driverService.get(1L);
+        driverService.delete(driver.getId());
+        driverService.update(new Driver("Volodymyrovych", "KK325564"));
```
maxline, line 35: how the DAO know which record should be  updated if we use the object without id  in method parameter?

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
```diff
@@ -22,8 +23,8 @@ public Manufacturer create(Manufacturer manufacturer) {
         try (Connection connection = ConnectionUtil.getConnection();
                 PreparedStatement statement
                         = connection.prepareStatement(query, Statement.RETURN_GENERATED_KEYS)) {
-            statement.setString(1, manufacturer.getName());
-            statement.setString(2, manufacturer.getCountry());
+            statement.setObject(1, manufacturer.getName());
+            statement.setObject(2, manufacturer.getCountry());
```
maxline, line 27: probably I missed something, why setObject() is better then setString()?

maxline (changes requested):
Good job! Just a couple of small comments, please take a look.

src/main/java/mate/jdbc/Main.java
copypasterr: @ArtemPonomar advised me to make it, perhaps this general method provides more flexibility and reduces the risk of errors

src/main/java/mate/jdbc/dao/impl/ManufacturerDaoImpl.java
copypasterr: @ArtemPonomar advised me to make it, perhaps this general method provides more flexibility and reduces the risk of errors

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        String query = "INSERT INTO drivers (name, license_number) "
+                + "VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement
+                        = connection.prepareStatement(query, Statement.RETURN_GENERATED_KEYS)) {
+            statement.setObject(1, driver.getName());
+            statement.setObject(2, driver.getLicenseNumber());
```
Andriy-dz, line 27: it's still better to use getString because you know the exact data type

Andriy-dz (approved):
Good Job!

PR #562, opened 2023-03-23
[HW-DAO-SERVICE] implement dao services for Driver and Manufacturer

marianapolishchuk2169:


src/main/resources/init_db.sql
```diff
@@ -7,3 +7,10 @@ CREATE TABLE `manufacturers` (
                                         `country` VARCHAR(225) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+                                        `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
+                                        `name` varchar(255) DEFAULT NULL,
```
sanya-1xx8, line 13: Variable types should be uppercase too

sanya-1xx8 (approved):
Good job!

src/main/resources/init_db.sql
maxline: Agree with previous comment. please rename in uppercase style

maxline (approved):
Well done! approved with one small comment, take a look for the future

PR #563, opened 2023-03-23
HomeWork/JDBC DAO vs Service

NickolayVoloshchuk:


src/main/resources/init_db.sql
```diff
@@ -7,3 +7,10 @@ CREATE TABLE `manufacturers` (
                                         `country` VARCHAR(225) NOT NULL,
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+                                 `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
+                                 `name` VARCHAR(225) NOT NULL,
+                                 `license_number` VARCHAR(225) NOT NULL,
+                                 `is_deleted` TINYINT NOT NULL DEFAULT 0,
+                                 PRIMARY KEY (`id`));
```
MaksymHorielikov, line 16: add an empty line to the end of the file

MaksymHorielikov (approved):

src/main/java/mate/jdbc/util/TestUtil.java
```diff
@@ -0,0 +1,110 @@
+package mate.jdbc.util;
+
+import java.util.Scanner;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
+public class TestUtil {
```
sarakhmen, line 9: Actually, it's a bad practice to mix your production code with test-related code. If you do such comprehensive testing, it's better to create separate test classes and use junit, mockito and so on

sarakhmen (changes requested):
Overall good job, but let's improve your solution :)

src/main/java/mate/jdbc/util/TestUtil.java
NickolayVoloshchuk: I know. But the README file says that we should test in the main (

sarakhmen: Consider moving it to the Main class than, instead of separate TestUtil

sarakhmen (approved):
Good job :)

PR #575, opened 2023-03-24
implemented all services

OlehSarapuk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,52 @@
...
+
+        Manufacturer bmw = new Manufacturer(1L, "BMW", "Germany");
+        System.out.println("create manufacturer");
+        System.out.println(manufactureService.create(bmw) + System.lineSeparator());
+        System.out.println("get manufacturer");
+        System.out.println(manufactureService.get(bmw.getId()) + System.lineSeparator());
+        List<Manufacturer> manufacturers = manufactureService.getAll();
+        System.out.println("getAll manufacturers");
+        manufacturers.stream()
+                .forEach(System.out::println);
```
outref, line 25:
```suggestion
        System.out.println("getAll manufacturers");
        manufactureService.getAll().stream()
                .forEach(System.out::println);
```
inline

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseGet(Manufacturer::new);
```
outref, line 22:
```suggestion
        return manufacturerDao.get(id);
```
i don't think you should return empty Manufacturer

outref (approved):
great!

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,49 @@
...
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufactureService manufactureService =
+                (ManufactureService) injector.getInstance(ManufactureService.class);
+
+        Manufacturer bmw = new Manufacturer(1L, "BMW", "Germany");
```
Andriy-dz, line 16:
```suggestion
        Manufacturer bmw = new Manufacturer(null, "BMW", "Germany");
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,49 @@
...
+        System.out.println(manufactureService.get(bmw.getId()) + System.lineSeparator());
+        System.out.println("getAll manufacturers");
+        manufactureService.getAll().stream()
+                .forEach(System.out::println);
+        System.out.println();
+        bmw.setName("MINI");
+        System.out.println("update manufacturer");
+        System.out.println(manufactureService.update(bmw) + System.lineSeparator());
+        System.out.println("delete manufacturer");
+        System.out.println(manufactureService.delete(bmw.getId()) + System.lineSeparator());
```
Andriy-dz, line 29: you can also display all drivers on the screen to show that the deletion works (Optional)

src/main/java/mate/jdbc/dao/DriverDaoImpl.java
```diff
@@ -0,0 +1,115 @@
...
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update a driver "
+                    + driver, e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE drivers SET is_deleted = TRUE WHERE id = ?";
```
Andriy-dz, line 98:
```suggestion
        String query = "UPDATE drivers SET is_deleted = TRUE WHERE id = ?;";
```

Andriy-dz (approved):
Very good solution! Aproved)

PR #576, opened 2023-03-24
hw-jv-dao-service

vika-trach:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
+        manufacturerService.create(new Manufacturer("Hyundai", "South Korea"));
+        driverService.create(new Driver("Petro","QW123456"));
+        driverService.create(new Driver("Vika","QS123656"));
+        driverService.create(new Driver("Bob","MW125456"));
+        System.out.println(driverService.get(1L));
+        System.out.println(driverService.get(2L));
+        System.out.println(driverService.get(3L));
+        System.out.println(manufacturerService.get(1L));
+        System.out.println(manufacturerService.get(2L));
+        System.out.println(manufacturerService.get(3L));
```
AlexandLes, line 28: It is better to use constant in this case

src/main/resources/init_db.sql
```diff
@@ -1,9 +1,19 @@
...
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
+
+CREATE TABLE `drivers` (
+  `id` int NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
+  `license_number` varchar(255) DEFAULT NULL,
+  `is_deleted` tinyint NOT NULL DEFAULT '0',
+  PRIMARY KEY (`id`)
```
AlexandLes, line 15: INT ,VARCHAR has to be in upper case.
And better to use BIGINT instead of INT

src/main/java/mate/jdbc/exeption/DataProcessingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.jdbc.exeption;
```
sarakhmen, line 1: Here's a minor typo in you package name:
```suggestion
package mate.jdbc.exception;
```

src/main/resources/init_db.sql
sarakhmen: agree with @AlexandLes. As stated in common mistakes to the previous HW - you should use bigint instead of int for id columns

sarakhmen (changes requested):
Overall very good job, but let's slightly improve your solution :)

sarakhmen (approved):
Good job ;)

PR #586, opened 2023-03-24
added solution

redmanO-o:
1. Class Driver - created;
2. ManufacturerService and implementation with overrided methods - created;
3. DriverService and implementation with overrided methods - created;
4. Naming rules for working with DB - followed;
5. Code in Main - tested;
6. Main contains only services;
7. Injection annotations - used;
8. init_db.sql - refreshed;
9. Maven checkstyle - BUILD SUCCESS;

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,57 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer mercedes = new Manufacturer("Mercedes", "Germany");
+        Manufacturer ferraris = new Manufacturer("Ferraris", "France");
+        Manufacturer lada = new Manufacturer("Lada", "USSR");
+
+        manufacturerService.create(mercedes);
+        manufacturerService.create(ferraris);
+        manufacturerService.create(lada);
+        manufacturerService.getAll().forEach(System.out::println);
+
+        Manufacturer ferrari = manufacturerService.get(ferraris.getId());
+        ferrari.setName("Ferrari");
+        ferrari.setCountry("Italy");
+        manufacturerService.update(ferrari);
+        System.out.println("updated" + manufacturerService.get(ferrari.getId()));
+        manufacturerService.getAll().forEach(manufacturer -> manufacturerService
+                .delete(manufacturer.getId()));
+        manufacturerService.getAll().forEach(System.out::println);
+
+        DriverService driverService
+                = (DriverService) injector.getInstance(DriverService.class);
+
+        Driver bob = new Driver("Bob", "123455");
+        Driver john = new Driver("John", "678909");
+        Driver alice = new Driver("ALice", "777777");
+
+        driverService.create(bob);
+        driverService.create(john);
+        driverService.create(alice);
+        driverService.getAll().forEach(System.out::println);
+
+        Driver johan = driverService.get(john.getId());
+        johan.setName("Johan");
+        johan.setLicenseNumber("999999");
+        driverService.update(johan);
+        System.out.println("updated" + driverService.get(johan.getId()));
+
+        driverService.getAll().forEach(driver -> driverService
+                .delete(driver.getId()));
+        driverService.getAll().forEach(System.out::println);
     }
+
 }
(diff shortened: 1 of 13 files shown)


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+
+    @Override
+    public Driver create(Driver driver) {
+        String query = "INSERT INTO drivers (name, driver_license) VALUES (?, ?)";
```
oleksiistepaniak, line 22: If you remember, Bohdan said that a rule of good tone is to add semicolon at the end of your query.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create driver. " + driver, e);
+        }
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        String query = "SELECT * FROM drivers"
+                + " WHERE id = ? AND is_deleted = FALSE";
```
oleksiistepaniak, line 42: Here is the same situation as above.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+            return Optional.ofNullable(driver);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get driver by id " + id, e);
+        }
+    }
+
+    @Override
+    public List<Driver> getAll() {
+        String query = "SELECT * FROM drivers WHERE is_deleted = FALSE";
```
oleksiistepaniak, line 59: Here is the same situation as above.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't get a list of drivers "
+                    + "from drivers table.", e);
+        }
+    }
+
+    @Override
+    public Driver update(Driver driver) {
+        String query = "UPDATE drivers SET name = ?, driver_license = ?"
+                + " WHERE id = ? AND is_deleted = FALSE";
```
oleksiistepaniak, line 78: Here is the same situation as above.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,112 @@
...
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't update a driver "
+                    + driver, e);
+        }
+    }
+
+    @Override
+    public boolean delete(Long id) {
+        String query = "UPDATE drivers SET is_deleted = TRUE WHERE id = ?";
```
oleksiistepaniak, line 95: Here is the same situation as above.

oleksiistepaniak (approved):
Overall, a perfect job! I approved your solution and I gave some tips on how to improve your solution.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
redmanO-o: I remember, but then the mentor said it's better not to do that. let's see how it will be this time))

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,56 @@
...
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer mercedes = new Manufacturer("Mercedes", "Germany");
+        Manufacturer ferraris = new Manufacturer("Ferraris", "France");
+        Manufacturer lada = new Manufacturer("Lada", "USSR");
```
petro-ushchuk, line 18: here we go again haha 😄

src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverDao {
+
```
petro-ushchuk, line 8: what should we do with redundant lines? please, check all the files

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
petro-ushchuk: both options are good, but without `;` it reads easier, since you have kinda `);";` and it is very noticeable

petro-ushchuk (approved):
Clean work 🔥

src/main/java/mate/jdbc/Main.java
redmanO-o: all for a good mood ;)

PR #591, opened 2023-03-25
hw-dao-service

katrevichdv:
Driver.class was created.
ManufacturerService and its implementation were created. 
DriverDao, DriverService and their implementations were created. pom.xml and init_db.sql were changed.
Main.main was implemented.


andriy-shulepa (approved):
Good jom

PR #597, opened 2023-04-07
created services for Manufacturer and Driver and made class Driver an…

vromanyuk:
…d DAO for Driver

### DIFF
--- a/src/main/java/mate/jdbc/model/Driver.java
+++ b/src/main/java/mate/jdbc/model/Driver.java
@@ -1,0 +1,70 @@
+package mate.jdbc.model;
+
+import java.util.Objects;
+
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver() {
+
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
+        this.id = id;
+        this.name = name;
+        this.licenseNumber = licenseNumber;
+    }
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
+        this.id = id;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public String getLicenseNumber() {
+        return licenseNumber;
+    }
+
+    public void setLicenseNumber(String licenseNumber) {
+        this.licenseNumber = licenseNumber;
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, licenseNumber);
+    }
+
+    @Override
+    public String toString() {
+        return "Manufacturer{"
```
olekskov, line 64: The class name in the `toString` method should be 'Driver' instead of 'Manufacturer'.

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
olekskov, line 21: Consider handling the case when `driverDao.get(id)` returns an empty Optional. You can use `orElseThrow` to throw a custom exception if the driver is not found. 🕵️

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
olekskov, line 21: Consider handling the case when Optional is empty, instead of calling .get() directly. You can use .orElseThrow() to throw a custom exception if the manufacturer is not found.

olekskov (changes requested):

src/main/java/mate/jdbc/model/Driver.java
vromanyuk: copy pust from Manufacturer :))

Ivan95kos (approved):
Great job

PR #598, opened 2023-04-09
initial commit

Rolik-Oleksandr:


src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id)
+                .orElseThrow(() -> new NoSuchElementException(
+                        "Not manufacturer found with index: " + id));
```
olekskov, line 25: Change the error message to 'No manufacturer found with index:' instead of 'Not manufacturer found with index:'. This will make the error message more clear and grammatically correct.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,111 @@
...
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get a list of drivers "
+                    + "from drivers table.", e);
+        }
+    }
+
+    @Override
+    public Driver update(Driver driver) {
+        String query = "UPDATE drivers SET name = ?, country = ?"
+                + " WHERE id = ? AND is_deleted = FALSE";
```
olekskov, line 77: The SQL query in the update method has a wrong column name 'country'. It should be 'license_number' instead.

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,52 @@
...
+
+    private static void deleteDriverById(Long id,
+                                         DriverService driverService) {
+        driverService.delete(id);
+    }
+
+    private static Optional<Driver> getDriverById(Long id,
+                                                  DriverService driverService) {
+        return Optional.ofNullable(driverService.get(id));
+    }
```
olekskov, line 34: DriverService already thrown exception if there isn't such id. Why we should cover result with Optional again?

olekskov (changes requested):

PR #610, opened 2023-05-03
Create class Driver, ManufacturerService and DriverService, and its i…

yulka-tsybulka:
…mplementation


andriy-shulepa (approved):
Good job

PR #614, opened 2023-05-13
HW-03

IhorPetrytskyi333:


Ivan95kos (changes requested):
make sure your build passes before requesting a review 😉

src/main/java/mate/jdbc/dao/DriverDao.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.dao;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverDao {
+
```
kozhukhovsky, line 8: remove empty line

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElse(null);
```
kozhukhovsky, line 21: it's better not to return null, in this case just get().get() is enough, and it's even better to throw an exception in the future
get().orElseThrow()

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
...
+
+    Manufacturer get(Long id);
+
+    List<Manufacturer> getAll();
+
+    Manufacturer update(Manufacturer manufacturer);
+
+    boolean delete(Long id);
+}
+
```
kozhukhovsky, line 17: delete empty line

kozhukhovsky (approved):

PR #617, opened 2023-05-16
Jv dao service b001

Poikka001:


src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+    Driver create(Driver driver);
+
+    Optional<Driver> get(Long id);
```
aaananas, line 10: you should not return an Optional from service layer, use orElseThrow

aaananas (changes requested):

src/main/java/mate/jdbc/service/DriverService.java
Rommelua: Not fixed

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
Rommelua, line 10: you should not return an Optional from service layer, use orElseThrow

Rommelua (changes requested):

Rommelua (approved):

PR #620, opened 2023-05-18
home-work/jv-dao-service

andrii-subash:


src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
olekskov, line 22: Use .orElseThrow() instead of .get() to handle NoSuchElementException.

src/main/java/mate/jdbc/model/Manufacturer.java
```diff
@@ -61,7 +66,7 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "Manufacturer{"
+        return "ManufacturerService{"
```
olekskov, line 69: The class name in the toString method should be 'Manufacturer' instead of 'ManufacturerService'.

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
olekskov, line 22: please use .orElseThrows construction with proper exception instead of raw get attempt

olekskov (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
kozhukhovsky, line 22: same

kozhukhovsky (approved):

PR #628, opened 2023-05-28
Created DB with tables manufacturers and drivers , implemented DAO an…

0-zagar-0:
…d SERVICE levels , checked in the main methodÑ

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,0 +1,128 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        String createDriverQuery = "INSERT INTO drivers(name, license_number) "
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+    
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+    
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
olekskov, line 22: Please use .orElseThrow() construction to rethrow exception with meaningful message in case driver wasn't found

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+    
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+    
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
olekskov, line 22: same here

olekskov (changes requested):

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    
+    private Driver getDriver(ResultSet resultSet) {
+        try {
+            Long id = resultSet.getObject("id", Long.class);
+            String name = resultSet.getString("name");
+            String licenseNumber = resultSet.getString("license_number");
+            return new Driver(id, name, licenseNumber);
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't get driver from DB", e);
+        }
```
Rommelua, line 126: use **throws**
```suggestion
            Long id = resultSet.getObject("id", Long.class);
            String name = resultSet.getString("name");
            String licenseNumber = resultSet.getString("license_number");
            return new Driver(id, name, licenseNumber);
```

Rommelua (approved):

PR #629, opened 2023-05-29
dao service solution first commit

ArtyomK75:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,85 @@
...
+                driverService, "Nick", "356894");
+        Driver copyDriverNick = driverService.get(driverNick.getId());
+        System.out.println(copyDriverNick);
+
+        Driver driverNancy = createRecordDriver(
+                driverService, "Nancy", "987562");
+        driverService.delete(driverNancy.getId());
+
+        List<Driver> driverList = driverService.getAll();
+        System.out.println(manufacturerList);
```
olekskov, line 66: Use 'driverList' instead of 'manufacturerList' when printing the list of drivers.

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
olekskov, line 22: Consider handling the case when manufacturerDao.get(id) returns an empty Optional. You can use .orElseThrow() with a custom exception to handle this scenario.

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, licenseNumber);
+    }
+
+    @Override
+    public String toString() {
+        return "Manufacturer{"
```
olekskov, line 65: The 'toString' method refers to the class as 'Manufacturer' instead of 'Driver'. Replace 'Manufacturer' with 'Driver'.

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
olekskov, line 22: You should check if the optional returned by driverDao.get(id) is present before calling get() on it. If the optional is empty, it means the requested driver is not found.
I'd recommend to use .orElseThrow method to produce proper exception with relevant message in such case

olekskov (changes requested):

Ivan95kos (approved):
Great job

PR #631, opened 2023-06-01
Create ManufacturerServiceImpl and DriverServiceImpl, DriverDaoImpl, …

Gamemod13:
…tested workability.

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,0 +1,114 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.exception.DataProcessingException;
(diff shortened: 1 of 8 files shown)


sZlishchev (changes requested):
Good job! Just add the init_db file

sZlishchev (approved):
Great job! Keep pushing)))

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
```
maxline, line 15: How the solution works if file `ConnectionUtil` with db connection parameters was not changed?
Looks like you forgot push the changes.

maxline (changes requested):
Well done! Left just one question.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
Gamemod13: I changed it, but I do not add it to VC because that class has a sensitive data (password)

PR #634, opened 2023-06-05
DriverDao implemented

Tupadzaky:
DriverService implemented
ManufacturerService implemented

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,64 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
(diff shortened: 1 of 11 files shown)


V-Solianyk (approved):
Great job, I don't know what can be improved here)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,64 @@
...
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        manufacturerService.create(new Manufacturer("Datsun", "Japan"));
+        manufacturerService.create(new Manufacturer("Toyota", "Japan"));
+
+        System.out.println("All manufacturers");
+        manufacturerService.getAll().forEach(System.out::println);
+        System.out.println(System.lineSeparator());
+
+        Manufacturer manufacturer = manufacturerService.get(1L);
```
maxline, line 23: Let's try to not hardcode `id` values, you can get needed `id` by using `manufacturer.getId()` method. 
Please fix all similar cases (not only for manufacturers) in this class.

maxline (changes requested):
Very good job! I left just one minor comment.

PR #635, opened 2023-06-05
created a new driver model and also divided the entire program into 3…

V-Solianyk:
… layers, tested the work in the main.

### DIFF
--- a/src/main/java/mate/jdbc/model/Driver.java
+++ b/src/main/java/mate/jdbc/model/Driver.java
@@ -2,23 +2,23 @@ package mate.jdbc.model;
 
 import java.util.Objects;
 
-public class Manufacturer {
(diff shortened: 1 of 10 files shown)


src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Objects;
+
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver(){
+    }
```
Tupadzaky, line 11: What is this constructor for?

Tupadzaky (approved):
I didn't find any problems, but check out the comment

sarakhmen (approved):
good job :)

PR #637, opened 2023-06-06
Manufacturer entity is a part of 3-tier arc.pattern as well as Driver…

TymurProkhorov:
…, all logic is called from service

### DIFF
--- a/src/main/resources/init_db.sql
+++ b/src/main/resources/init_db.sql
@@ -2,16 +2,17 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
 USE `taxi_service`;
 
 CREATE TABLE `manufacturers` (
-                                        `id` BIGINT(11) NOT NULL AUTO_INCREMENT,
-                                        `name` VARCHAR(225) NOT NULL,
-                                        `country` VARCHAR(225) NOT NULL,
-                                        `is_deleted` TINYINT NOT NULL DEFAULT 0,
-                                        PRIMARY KEY (`id`));
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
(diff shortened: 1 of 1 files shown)


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
...
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService =
```
overpathz, line 11: You don't test your ManufacturerService

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,25 @@
 package mate.jdbc;
```
overpathz, line 1: There is no "init_db.sql". Add to PR this file.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create driver. " + driver, e);
+        }
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        String query = "SELECT * FROM drivers"
+                + " WHERE id = ? AND is_deleted = 'FALSE';";
```
overpathz, line 43: Here, 'FALSE' is treated as a string literal.
Use just FALSE (and in other places in the entire code).

overpathz (changes requested):
Overall very good.
Please, fix little comments to approve

aaananas (approved):

PR #638, opened 2023-06-08
implemented task

chrisnyrka:


marimarig (approved):
Good job:)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,36 @@
...
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+
+        Driver driver = new Driver();
+        driver.setName("Moris");
+        driver.setLicenseNumber("9345");
+        System.out.println(driverService.create(driver));
+        System.out.println(driverService.get(1L));
+        System.out.println(driverService.delete(2L));
```
sarakhmen, line 20: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

sarakhmen (changes requested):
Overall excellent job! You can slightly adjust your main method code if you want ;)

sarakhmen (approved):

PR #639, opened 2023-06-08
Hw solution

ykhmaruk:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
+    }
```
Pkrasota, line 23: please use orElseThrow() for managing case where Optional is empty, e.g.
return driverDao.get(id).orElseThrow(() -> new NoSuchElementException("Could not get driver. ID:" + id));

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
+    }
+
```
Pkrasota, line 24: And here also orElseThrow()

Pkrasota (changes requested):
Good job, some minor mistake

Pkrasota (approved):
Well done :)

src/main/resources/init_db.sql
```diff
@@ -2,8 +2,17 @@ CREATE SCHEMA IF NOT EXISTS `taxi_service` DEFAULT CHARACTER SET utf8;
...
+  PRIMARY KEY (`id`));
+
+CREATE TABLE `drivers` (
+  `id` bigint NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) DEFAULT NULL,
+  `license_number` varchar(255) DEFAULT NULL,
+  `is_deleted` tinyint DEFAULT '0',
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3;
+
```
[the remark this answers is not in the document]
sarakhmen: Agree with @Pkrasota ;)

sarakhmen (approved):
LGTM! You can improve your solution if you want :)

PR #640, opened 2023-06-08
Hw 1

marimarig:


chrisnyrka (approved):
Great work :)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,37 @@
...
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setName("Audi");
+        manufacturer.setCountry("Germany");
+        Manufacturer toyota = new Manufacturer(3L, "Toyota", "Japan");
+        System.out.println(manufacturerService.create(manufacturer));
+        System.out.println(manufacturerService.get(manufacturer.getId()));
+        System.out.println(manufacturerService.delete(toyota.getId()));
+        System.out.println(manufacturerService.update(new Manufacturer(3L, "BMW","Germany")));
```
sarakhmen, line 23: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

sarakhmen (approved):
LGTM! But you can refactor/improve your main method's code if you want :)

PR #642, opened 2023-06-10
add dao-service solution

AnastasiiaVariuchuk:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver tempDriver = new Driver("Ihor", "11111");
+        Driver artur = new Driver("Artur", "11112");
+        driverService.create(tempDriver);
+        tempDriver.setName("Nikita");
+        tempDriver.setLicenseNumber("11113");
+        driverService.update(tempDriver);
+        System.out.println(driverService.get(tempDriver.getId()));
+        driverService.delete(tempDriver.getId());
+        System.out.println(driverService.getAll());
```
m-pnsk, line 36: it will be better if you move this logic to two different functions

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElseThrow(
+                () -> new NoSuchElementException("Could not get driver. ID:" + id));
+    }
```
m-pnsk, line 25: Why should the Manufacturer get the drivers?
```suggestion
    public Manufacturer get(Long id) {
        return manufacturerDao.get(id).orElseThrow(
                () -> new NoSuchElementException("Could not get manufacturer. ID:" + id));
    }
```

m-pnsk (changes requested):

src/main/java/mate/jdbc/Main.java
AnastasiiaVariuchuk: good idea)

m-pnsk (approved):

PR #644, opened 2023-06-11
method main() works, checkstyle done

MykNich:


Poloutinho (approved):
Good job!

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,47 @@
...
+
+        Manufacturer manufacturer = new Manufacturer("Chevrolet", "USA");
+        System.out.println(manufacturerService.create(manufacturer));
+
+        manufacturer.setCountry("DAEWOO-ZAZ");
+        manufacturer.setName("Ukraine");
+        System.out.println(manufacturerService.update(manufacturer));
+
+        System.out.println(manufacturerService.delete(6L));
+        System.out.println(manufacturerService.get(1L));
```
sarakhmen, line 26: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding method

src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.model.Driver;
+
+public interface DriverService {
+
```
sarakhmen, line 7:
```suggestion
```

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+
+    Manufacturer create(Manufacturer manufacturer);
+
```
sarakhmen, line 9:
```suggestion
```

sarakhmen (changes requested):
Good job, but let's improve your solution :) Please also add corresponding changes to your src/main/resources/init_db.sql script

sarakhmen (approved):
Good job :)

PR #646, opened 2023-06-11
Solved solution, with using services

Pkrasota:


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Optional;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.exception.DataProcessingException;
+import mate.jdbc.lib.Dao;
+import mate.jdbc.model.Driver;
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+    private static final String INSERT_FORMAT_REQUEST =
```
ykhmaruk, line 19: 'request' is more related to the HTTP request, so let's name your queries with 'query' suffix. Change it in all places

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,119 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement createStatement =
+                        connection.prepareStatement(INSERT_FORMAT_REQUEST,
+                             Statement.RETURN_GENERATED_KEYS)) {
+            createStatement.setString(1, driver.getName());
+            createStatement.setString(2, driver.getLicenseNumber());
+            createStatement.executeUpdate();
+            ResultSet generatedKeys = createStatement.getGeneratedKeys();
+            if (generatedKeys.next()) {
+                Long id = generatedKeys.getLong(1);
```
ykhmaruk, line 41: we can avoid creating additional variable here

ykhmaruk (commented):
Good job! Just a few recommendations

ykhmaruk (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,34 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+
```
sarakhmen, line 10:
```suggestion
```

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,34 @@
...
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        DriverService driverService =
+                (DriverService) injector.getInstance(DriverService.class);
+        Driver driver = new Driver("Alex", "51252");
+        System.out.println(driverService.create(driver));
+        System.out.println(driverService.get(1L));
+        System.out.println(driverService.delete(2L));
```
sarakhmen, line 19: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

sarakhmen (approved):
LGTM! You can improve your soluiton if you want ;)

PR #647, opened 2023-06-11
Service level has been implemented

m-pnsk:


src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver() {
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
```
AnastasiiaVariuchuk, line 11: you can create one more constructor here
```
public Driver(String name, String licenseNumber) {
        this.name = name;
        this.licenseNumber = licenseNumber;
    }
```

AnastasiiaVariuchuk: then you can create driver in this way
`Driver tempDriver = new Driver("Ihor", "11111");`

AnastasiiaVariuchuk: but it's just my recommendation;)

AnastasiiaVariuchuk: you also can create such constructor in Manufacturer)

AnastasiiaVariuchuk (commented):
Good job)

src/main/java/mate/jdbc/model/Driver.java
m-pnsk: Yes, but the setters do their job well :)

AnastasiiaVariuchuk (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,75 @@
...
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        System.out.println("Manufacturer Service testing:");
+        exampleOfUsageManufacturerService(manufacturerService);
```
rovein, line 17: Remember about method naming convention.
```suggestion
        runExampleOfUsageManufacturerService(manufacturerService);
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import mate.jdbc.util.ConnectionUtil;
+
+@Dao
+public class DriverDaoImpl implements DriverDao {
+    @Override
+    public Driver create(Driver driver) {
+        String addToDriversTableQuery = "INSERT INTO drivers (name, license_number) "
+                + "VALUES (?, ?);";
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement addToDriversTableStatement
```
rovein, line 24: Naming is good, but for simplicity you could also use simply:
```suggestion
                PreparedStatement statement

```
If you are working with one statement within one method.
**It is not an error, just my personal recommendation. Fix is optional**

rovein (approved):
Great job! See minor comments

PR #648, opened 2023-06-11
add dao layer to my solution

ArinaPolna:


src/main/java/mate/jdbc/service/DriverService.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Driver;
+
+@Service
```
AnastasiiaVariuchuk, line 7: it is sufficient to use this annotation in the implementation

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public interface ManufacturerService {
```
AnastasiiaVariuchuk, line 8: same stuff

AnastasiiaVariuchuk (commented):
Good job, just some comments))

AnastasiiaVariuchuk (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,50 @@
...
+        Manufacturer reno = new Manufacturer("reno", "France");
+        manufacturerService.create(audi);
+        manufacturerService.create(volvo);
+        manufacturerService.create(reno);
+        System.out.println(manufacturerService.get(audi.getId()));
+        manufacturerService.delete(audi.getId());
+        Manufacturer manufacturerToUpdate = new Manufacturer(2L, "shkoda", "Japan");
+        manufacturerService.update(manufacturerToUpdate);
+        Manufacturer manufacturerWithIndex3 = manufacturerService.get(3L);
+        Manufacturer manufacturerToCompare = new Manufacturer(3L, "reno", "France");
```
sarakhmen, line 27: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

sarakhmen (approved):
LGTM! You can improve your solution if you want :)

Nick97-git (approved):

PR #649, opened 2023-06-12
added first version of solution for task dao-service

Fiasko3304:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driver = new Driver();
+        driver.setName("Alex");
+        driver.setLicenseNumber("1499");
+        driverService.create(driver);
+        driver.setName("Artem");
+        driver.setLicenseNumber("0982");
+        driverService.update(driver);
+        driverService.get(2L);
+        driverService.delete(1L);
```
Pkrasota, line 33: And here also

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = new Manufacturer("BMW","Germany");
+        manufacturerService.create(manufacturer);
+        manufacturer.setName("Nissan");
+        manufacturer.setCountry("Japan");
+        manufacturerService.update(manufacturer);
+        manufacturerService.get(1L);
+        manufacturerService.delete(1L);
```
Pkrasota, line 22: here in your example/test code let's not use hardcoded int literals as id parameter. I'd recommend creating and assigning the manufacturer to some variable first. And then you can retrieve from this variable an id field and pass it to the corresponding methods

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,38 @@
...
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        ManufacturerService manufacturerService
+                = (ManufacturerService) injector.getInstance(ManufacturerService.class);
+
+        Manufacturer manufacturer = new Manufacturer("BMW","Germany");
```
Pkrasota, line 16: redant empty line

Pkrasota (approved):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,37 @@
...
+        manufacturer.setName("Nissan");
+        manufacturer.setCountry("Japan");
+        Manufacturer manufacturerId = manufacturerService.update(manufacturer);
+        manufacturerService.get(manufacturerId.getId());
+        manufacturerService.delete(manufacturerId.getId());
+
+        DriverService driverService = (DriverService) injector.getInstance(DriverService.class);
+        Driver driver = new Driver();
+        driver.setName("Alex");
+        driver.setLicenseNumber("1499");
```
maxline, line 26: You can also create constructor with 2 parameters and use it instead of using default constructor with setters.

maxline (approved):
Well done! Approved with small comment. Please take a look for the future

PR #652, opened 2023-06-12
Added hw solution

pkobevko:


ViktorBykh (changes requested):
Everything looks great)
I have a few suggestions

ViktorBykh (approved):
Everything looks great)

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,42 @@
...
+        driver.setLicenseNumber("7777777");
+        driverService.create(driver);
+        Driver driver1 = driverService.get(1L);
+        System.out.println(driver1);
+        driverService.getAll().forEach(System.out::println);
+        driver.setName("Petro Poroshenko Chocolate King");
+        driverService.update(driver);
+        driverService.getAll().forEach(System.out::println);
+        driverService.delete(1L);
+        driverService.getAll().forEach(System.out::println);
```
rovein, line 26: It's better to separate logically grouped lines of code by empty blocks for the sake of readability and easier understanding either in the future or by other devs.

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow(() ->
+                new RuntimeException("There is no such drivers with id: " + id));
```
rovein, line 23:
```suggestion
                new NoSuchElementException("There is no such drivers with id: " + id));
```

rovein (approved):
Good job! See some minor comments.

PR #654, opened 2023-06-12
Added Driver model, DriverDao and DriverService plus ManufacturerService

viprko:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,57 @@
...
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+        List<Driver> driversList = new ArrayList<>();
+        driversList.add(new Driver("Oleksii", "AST 445823"));
```
volodymyrruso, line 16: you can initialize it by using `        Driver volodymyr = driverService.create(new Driver("Oleksii", "AST 445823"));
                                                       driverService.create(Oleksii);
` but i'm not sure if my version is better, let's mentor decide 😃

volodymyrruso (approved):
good job, and very interesting work, i left some comments but i'm not sure if they're right I let the mentor decides if my comments are valid

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,57 @@
...
+        System.out.println("Driver with id: " + updateDriver.getId() + " after update = "
+                + driverService.get(3L));
+
+        driverService.delete(3L);
+        System.out.println("Drivers table after deletion driver with id 3 = "
+                + driverService.getAll());
+
+        ManufacturerService manufacturerService =
+                (ManufacturerService) injector.getInstance(ManufacturerService.class);
+        List<Manufacturer> manufacturersList = new ArrayList<>();
```
maxline, line 36:
```suggestion
        List<Manufacturer> manufacturers = new ArrayList<>();
```

maxline (approved):
Well done! Approved with small comment. Please take a look for the future

PR #660, opened 2023-06-12
build 0.9 std rev

maksigors:


src/main/java/mate/jdbc/util/ConnectionUtil.java
```diff
@@ -6,23 +6,27 @@
...
     static {
         try {
             Class.forName(JDBC_DRIVER);
         } catch (ClassNotFoundException e) {
-            throw new RuntimeException("Can't find SQL Driver", e);
+            throw new RuntimeException("Can't find Driver for SQL", e);
         }
     }
 
     public static Connection getConnection() throws SQLException {
```
DV7Art, line 22: i think if you throws in this block you dont need to use try-catch. Cause an exception declared with throws must be handled in the calling method, which calls the method that raised the exception.

DV7Art:
generally good, but I think it would be nice to repeat the exceptions

DV7Art (approved):

src/main/java/mate/jdbc/util/ConnectionUtil.java
maksigors: I don't think so)

sarakhmen (approved):
good job, you can improve your solution if you want ;)

PR #661, opened 2023-06-12
dao service hw solution

nechupara:


src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).orElse(null);
```
Fafakepa, line 22: I'm not sure, but it's probably better to throw an exception here

Fafakepa (changes requested):
Overall good job, but you forgot init_db.sql file and add @Dao annotation to ManufacturerDaoImpl

sarakhmen (approved):
Overall good job, you can improve your solution if you want ;)

PR #663, opened 2023-06-13
Solution dao_service

MTM1963:


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,108 @@
...
+            statement.setString(1, driver.getName());
+            statement.setString(2, driver.getLicenseNumber());
+            statement.executeUpdate();
+            ResultSet resultSet = statement.getGeneratedKeys();
+            if (resultSet.next()) {
+                driver.setId(resultSet.getObject(1, Long.class));
+            }
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't create manufacturer. " + driver, e);
```
kozhukhovsky, line 34: Change the error message to 'Couldn't create driver.' instead of 'Couldn't create manufacturer.'

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
```
kozhukhovsky, line 22: Consider handling the Optional in the get method and throwing an exception if the driver is not found.

src/main/java/mate/jdbc/service/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
```
kozhukhovsky, line 21: Consider returning Driver instead of Optional<Driver> to match the interface definition.

src/main/java/mate/jdbc/model/Driver.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class Driver {
+    private Long id;
+    private String name;
+    private String licenseNumber;
+
+    public Driver() {
+    }
+
+    public Driver(Long id, String name, String licenseNumber) {
```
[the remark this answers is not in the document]
MTM1963: How should I do that? Сause, in DriverDaoImpl, I created method getDriver which return new Driver(id, name, licenseNumber); So, I don't understand, how I could remove id from constructor. If I remove it, it will be null id.

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
MTM1963: done.

src/main/java/mate/jdbc/service/DriverServiceImpl.java
MTM1963: done.

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    @Inject
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
```
aaananas, line 21: you should not return optional from service layer

aaananas (changes requested):

src/main/java/mate/jdbc/service/ManufacturerServiceImpl.java
MTM1963: Done.

PR #667, opened 2023-06-17
Driver and Manufacturer Model + Dao + Service created.

Hlib13:
(Need explanation. I need create a new date base for this project ? If yes,i have problem with console connection to idea)

### DIFF
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -1,7 +1,20 @@
 package mate.jdbc;
 
+import mate.jdbc.lib.Injector;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
 public class Main {
+    private static Injector injector = Injector.getInstance("mate.jdbc");
+
     public static void main(String[] args) {
-        // test your code here
+
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        manufacturerService.getAll().forEach(System.out::println);
+
+        DriverService driverService = (DriverService)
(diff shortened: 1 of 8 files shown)


Ivan95kos (changes requested):
Ask this question in the Q&A

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
okuzan, line 11: should return Manufacturer, it's service not dao

okuzan (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,20 @@
...
     public static void main(String[] args) {
-        // test your code here
+
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        manufacturerService.getAll().forEach(System.out::println);
+
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        driverService.getAll().forEach(System.out::println);
```
olekskov, line 18: Please use main class to properly test project by invocation of all service layer methods.
Now you're just getting all drivers without adding any data, don't think you can get anything except empty list

olekskov (changes requested):

aaananas (approved):

PR #668, opened 2023-06-26
solution hw dao+service

krismva:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
olekskov, line 22: in such cases it's better to add some meaningful exception with proper message

olekskov (approved):
Good job overall, approved.
Please check comment below for further tasks

PR #670, opened 2023-06-27
Implemented driver / manufacturer services

vaniii1:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
kozhukhovsky, line 22: if we have an empty value here, there will be an error
It is better not to use get().get(), but get().orElseThrow() and throw an exception with a clear message

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
kozhukhovsky, line 22: same

kozhukhovsky (changes requested):

Ivan95kos (approved):
Nice

PR #674, opened 2023-07-06
realization jv-jbdc homwork

bigvini:


src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    private ManufacturerDao manufacturerDao = new ManufacturerDaoImpl();
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
okuzan, line 23: use orElseThrow

okuzan (changes requested):

Ivan95kos (approved):
Great job

PR #677, opened 2023-07-12
1. Created class Driver

sshestaka:
2. Created DriverDao interface and implementation.
3. Created ManufacturerService and DriverService, and their implementations
4. DB connected
5. In Main class created tests.
6. In DaoImplementations in the UPDATE methods, checks have been added for the execution of the request - successfully or not.
7. In the Manufacturer and Driver models one more constructor has been added without entering "id".

### DIFF
--- a/src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
+++ b/src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
@@ -1,0 +1,39 @@
+package mate.jdbc.service.impl;
+
+import java.util.List;
+import mate.jdbc.dao.DriverDao;
+import mate.jdbc.lib.Inject;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Driver;
+import mate.jdbc.service.DriverService;
+
+@Service
+public class DriverServiceImpl implements DriverService {
+    @Inject
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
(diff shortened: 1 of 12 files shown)


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).get();
```
okuzan, line 22: orElseThrow Nosuchelement with message with id

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao manufacturerDao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id).get();
```
okuzan, line 22: same

okuzan (changes requested):

Ivan95kos (approved):
Great job

PR #678, opened 2023-07-13
Solution

OlegMR0:


src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,21 @@
...
-        // test your code here
+        Driver driver = new Driver("bob", "number");
+        DriverDao driverDao = (DriverDao) injector.getInstance(DriverDao.class);
+        driver = driverDao.create(driver);
+        driver.setName("Bobchello");
+        driverDao.update(driver);
+        driverDao.getAll().forEach(System.out::println);
+        System.out.println(driverDao.get(driver.getId()));
+        System.out.println(driverDao.delete(driver.getId()));
+
```
kozhukhovsky, line 19:
```suggestion
```

src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,116 @@
...
+                          Query.INSERT.string, Statement.RETURN_GENERATED_KEYS)) {
+            statement.setString(1, driver.getName());
+            statement.setString(2, driver.getLicenseNumber());
+            statement.executeUpdate();
+            ResultSet generatedKeys = statement.getGeneratedKeys();
+            generatedKeys.next();
+            driver.setId(generatedKeys.getObject(1, Long.class));
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Can't insert driver. " + driver.toString(), e);
```
kozhukhovsky, line 32:
```suggestion
            throw new DataProcessingException("Can't insert driver. " + driver, e);
            
```

When an object is cast to a String, whether it's the print method or string concatenation, toString doesn't need to be called explicitly, it's called by itself. Fix this everywhere, please.

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao dao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return dao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return dao.get(id).get();
```
kozhukhovsky, line 22:
```suggestion
        return dao.get(id).orElseThrow(...);

```

if optional is empty, there will be an error.
In such places, you need to do .orElseThrow() and throw some kind of exception with a clear message.

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private ManufacturerDao dao;
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return dao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return dao.get(id).get();
```
kozhukhovsky, line 22: same

src/main/resources/init_db.sql
```diff
@@ -7,3 +7,11 @@ CREATE TABLE `manufacturers` (
...
                                         `is_deleted` TINYINT NOT NULL DEFAULT 0,
                                         PRIMARY KEY (`id`));
+CREATE TABLE `drivers` (
+                           `id` bigint NOT NULL AUTO_INCREMENT,
+                           `name` varchar(45) NOT NULL,
+                           `license_number` varchar(45) NOT NULL,
+                           `is_deleted` tinyint NOT NULL DEFAULT '0',
+                           PRIMARY KEY (`id`)
+)
+
```
kozhukhovsky, line 17:
```suggestion
```

kozhukhovsky (changes requested):

Ivan95kos (approved):
Great job

PR #680, opened 2023-07-17
added solution

Serhii-Nosulia:


kozhukhovsky (changes requested):
Good job!

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow();
```
kozhukhovsky, line 22: It's almost there, you just need to throw some kind of exception with a clear message :)

kozhukhovsky (approved):

PR #685, opened 2023-07-19
Create class Driver, create interface ManufacturerService and all CRU…

TemkaHillfinger:
…D methods ,create interface DriverService  and all CRUD methods

### DIFF
--- a/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
+++ b/src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
@@ -1,0 +1,113 @@
+package mate.jdbc.dao.impl;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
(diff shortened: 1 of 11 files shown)


src/main/java/mate/jdbc/dao/impl/DriverDaoImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        try (Connection connection = ConnectionUtil.getConnection();
+                PreparedStatement statement
+                        = connection.prepareStatement(query)) {
+            statement.setString(1, driver.getName());
+            statement.setString(2,driver.getLicenseNumber());
+            statement.setLong(3, driver.getId());
+            statement.executeUpdate();
+            return driver;
+        } catch (SQLException e) {
+            throw new DataProcessingException("Couldn't delete a driver by id ", e);
```
olekskov, line 89: Method is "update" but exception contains "delete"

olekskov (approved):
Good job

PR #686, opened 2023-07-19
Completed task

AntonU91:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id)
+                        .orElse(null);
```
olekskov, line 23: Generally it's not a good practice to return null. It may easily cause nullpointer exception further in code.
Please use orElseThrow() method instead. With such approach you will definitely know where problem is and avoid debugging to determine problem cause

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        return manufacturerDao.get(id)
+                              .orElse(null);
+    }
```
olekskov, line 24: same as above

olekskov (changes requested):

src/main/java/mate/jdbc/Main.java
```diff
@@ -1,7 +1,46 @@
...
-        // test your code here
+        ManufacturerService manufacturerService =
+                (ManufacturerService) INJECTOR.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer();
+        manufacturer.setCountry("Japan");
+        manufacturer.setName("Toyota");
+
+        Manufacturer createdManufacture = manufacturerService.create(manufacturer);
+
+        System.out.println(manufacturerService.get(29L));
```
Ivan95kos, line 21: For the future, don't hardcode ids, retrieve by getId()

Ivan95kos (approved):
Great job, with minor comments

PR #687, opened 2023-07-20
add my solution

mykhailoKrlKrk:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return driverDao.get(id);
+    }
```
olekskov, line 24: according to task you should return model, not Optional
- Driver get(Long id);

Please use orElseThrow() method with proper exception in case model is absent while getting model from optional

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Optional<Manufacturer> get(Long id) {
+        return manufacturerDao.get(id);
+    }
```
olekskov, line 24: same as above and do not forget to make related changes to both interfaces

olekskov (changes requested):

Ivan95kos (approved):
Nice

PR #690, opened 2023-07-22
jv-dao-service

s-hrynus:
…lized n-tier architecture

### DIFF
--- a/src/main/java/mate/jdbc/Application.java
+++ b/src/main/java/mate/jdbc/Application.java
@@ -1,0 +1,35 @@
+package mate.jdbc;
+
+import mate.jdbc.lib.Injector;
+import mate.jdbc.model.Driver;
+import mate.jdbc.model.Manufacturer;
+import mate.jdbc.service.DriverService;
+import mate.jdbc.service.ManufacturerService;
+
+public class Application {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void run() {
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver("Jane", "qwerty007");
+        driverService.create(driver);
+        driverService.getAll();
+        driverService.get(driver.getId());
+        driver.setName("Mike");
+        driver.setLicenseNumber("qwerty003");
+        driverService.update(driver);
+        driverService.delete(driver.getId());
+
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer("Volvo", "Germany");
+        manufacturerService.create(manufacturer);
+        manufacturerService.getAll();
+        manufacturerService.get(manufacturer.getId());
+        manufacturer.setName("Lexus");
+        manufacturer.setCountry("Japan");
+        manufacturerService.update(manufacturer);
+        manufacturerService.delete(manufacturer.getId());
+    }
+}
--- a/src/main/java/mate/jdbc/Main.java
+++ b/src/main/java/mate/jdbc/Main.java
@@ -2,6 +2,6 @@ package mate.jdbc;
 
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Application.run();
     }
 }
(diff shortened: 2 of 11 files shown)


src/main/java/mate/jdbc/Application.java
```diff
@@ -0,0 +1,35 @@
...
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer("Volvo", "Germany");
+        manufacturerService.create(manufacturer);
+        manufacturerService.getAll();
+        manufacturerService.get(manufacturer.getId());
+        manufacturer.setName("Lexus");
+        manufacturer.setCountry("Japan");
+        manufacturerService.update(manufacturer);
+        manufacturerService.delete(manufacturer.getId());
```
olekskov, line 33: You updated and deleted manufacturer but haven't used any get method to check if changes were applied

src/main/java/mate/jdbc/Main.java
```diff
@@ -2,6 +2,6 @@
 
 public class Main {
     public static void main(String[] args) {
-        // test your code here
+        Application.run();
     }
```
olekskov, line 6: Why do we need additional class with method run() just to call it from main?
Isn't it easier to just place code in main method?

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        Optional<Driver> driver = driverDao.get(id);
+        return driver.get();
+    }
```
olekskov, line 25: Please use .orElseThrow() method with proper exception for the case dao returned empty optional

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Manufacturer create(Manufacturer manufacturer) {
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer get(Long id) {
+        Optional<Manufacturer> manufacturer = manufacturerDao.get(id);
+        return manufacturer.get();
+    }
```
olekskov, line 25: same here

olekskov (changes requested):

src/main/java/mate/jdbc/Main.java
s-hrynus: I did it because I think If it's an application, it's better to create another class and put their code. 
As a result, I call run() in main.

src/main/java/mate/jdbc/Application.java
```diff
@@ -0,0 +1,35 @@
...
+        driver.setLicenseNumber("qwerty003");
+        driverService.update(driver);
+        driverService.delete(driver.getId());
+
+        ManufacturerService manufacturerService = (ManufacturerService)
+                injector.getInstance(ManufacturerService.class);
+        Manufacturer manufacturer = new Manufacturer("Volvo", "Germany");
+        manufacturerService.create(manufacturer);
+        manufacturerService.getAll();
+        manufacturerService.get(manufacturer.getId());
```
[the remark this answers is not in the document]
s-hrynus: @olekskov I used "sout" and checked it in the table.
So should I need to add "sout" here for print?

src/main/java/mate/jdbc/Application.java
```diff
@@ -0,0 +1,35 @@
...
+public class Application {
+    private static final Injector injector = Injector.getInstance("mate.jdbc");
+
+    public static void run() {
+        DriverService driverService = (DriverService)
+                injector.getInstance(DriverService.class);
+        Driver driver = new Driver("Jane", "qwerty007");
+        driverService.create(driver);
+        driverService.getAll();
+        driverService.get(driver.getId());
```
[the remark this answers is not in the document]
s-hrynus: same like I wrote bellow

olekskov (approved):
Good job

src/main/java/mate/jdbc/Application.java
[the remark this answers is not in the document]
olekskov: Understood, but it's better to have such steps in code so mentor can see flow of manual testing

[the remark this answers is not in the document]
s-hrynus: Roger that, thank you)

PR #693, opened 2023-07-24
Solution of this task

IvanovYurii:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+    private DriverDao driverDao;
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Optional<Driver> get(Long id) {
+        return Optional.ofNullable(driverDao.get(id).get());
```
okuzan, line 23: orElseThrow

okuzan (changes requested):

src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.lib.Service;
+import mate.jdbc.model.Manufacturer;
+
+@Service
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
okuzan, line 12: dao should return optional for get, but not service

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public Driver get(Long id) {
+        if (driverDao.get(id).isPresent()) {
+            return driverDao.get(id).get();
+        } else {
+            throw new DataProcessingException("Driver with ID " + id + " not found.",
+                    new RuntimeException("Driver not found in DB."));
+        }
```
okuzan, line 28: we have built-in functionality for that, use get(id).orElseThrow(() -> new ..

okuzan (changes requested):

olekskov (approved):

PR #697, opened 2023-08-01
Homework jv-dao-service.

UniyatYura:


src/main/java/mate/jdbc/service/ManufacturerService.java
```diff
@@ -0,0 +1,17 @@
+package mate.jdbc.service;
+
+import java.util.List;
+import java.util.Optional;
+import mate.jdbc.model.Manufacturer;
+
+public interface ManufacturerService {
+    Manufacturer create(Manufacturer manufacturer);
+
+    Optional<Manufacturer> get(Long id);
```
okuzan, line 10: get on service layer should return actual object, not optional

okuzan (changes requested):

src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Override
+    public Driver create(Driver driver) {
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver get(Long id) {
+        return driverDao.get(id).orElseThrow(() ->
+                new RuntimeException("Can't get driver from DB with id: " + id));
```
okuzan, line 23:
```suggestion
                new NoSuchElementException("Can't get driver from DB with id: " + id));
```

okuzan (approved):

PR #698, opened 2023-08-01
Solution

Volodymyr-Mykychak:


src/main/java/mate/jdbc/service/impl/DriverServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Override
+    public Driver createDriver(Driver driver) {
+        validationService.validateBeforeCreate(driver);
+        return driverDao.create(driver);
+    }
+
+    @Override
+    public Driver getDriverById(Long id) {
+        return driverDao.get(id).orElse(null);
```
okuzan, line 30:
```suggestion
        return driverDao.get(id).orElse(null);
```
better practice is to throw exception than to return null silently
```suggestion
        return driverDao.get(id).orElseThrow(() -> new NoSuchElementException("Couldn't find a driver with id: " + id));
```

src/main/java/mate/jdbc/service/impl/ManufacturerServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public Manufacturer createManufacturer(Manufacturer manufacturer) {
+        validationService.validateBeforeCreate(manufacturer);
+        return manufacturerDao.create(manufacturer);
+    }
+
+    @Override
+    public Manufacturer getManufacturerById(Long id) {
+        return manufacturerDao.get(id).orElse(null);
```
okuzan, line 31: same

okuzan (changes requested):

olekskov (approved):
Good job

